PR #83, opened 2017-09-21
Accommodate DUO ontology import

rushtong:
## Addresses:
https://broadinstitute.atlassian.net/browse/GAWB-2523
Requires: https://github.com/broadinstitute/consent/pull/244
Requires: https://github.com/broadinstitute/consent-data-use/pull/6

## Changes:
* Upgrade the owl api to handle direct ontology imports
* Update the app so that it doesn't fail on startup if it cannot find ontology files in GCS. Instead, log an error and re-check before trying to use any GCS ontology file.
* Use DUO terms where required instead of DUOS terms. 

## Testing:
* See the testing section in https://github.com/broadinstitute/consent/pull/244
* After updating the elastic search index with https://github.com/broadinstitute/consent/pull/244, the auto-complete feature should find DUO terms such as `methods research` and `non profit` as well as DUOS terms such as `male` and `female`
* Matching should still work even with DUO terms. In addition to current unit tests (which do use some DUO terms), here are some examples:
```
DUO_0000015 --> methods research
DOID_4      --> disease
Should succeed:
{
	"purpose":{"type":"named","name":"http://purl.obolibrary.org/obo/DUO_0000015"},
	"consent":{"type":"everything"}
}
Should succeed:
{
	"purpose":{"type":"named","name":"http://purl.obolibrary.org/obo/DUO_0000015"},
	"consent":{"type":"named","name":"http://purl.obolibrary.org/obo/DUO_0000015"}
}
Should fail:
{
	"purpose":{"type":"named","name":"http://purl.obolibrary.org/obo/DOID_4"},
	"consent":{"type":"named","name":"http://purl.obolibrary.org/obo/DUO_0000015"}
}
```

---

- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: JIRA ticket checks:
- [x] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [x] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [x] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [x] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [x] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Assign to submitter** to finalize
- [x] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Squash commits and merge to develop
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [x] **Submitter**: Verify swagger UI on dev environment still works after deployment
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
@@ -15,10 +15,16 @@ import java.net.URL;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static org.broadinstitute.dsde.consent.ontology.datause.builder.UseRestrictionBuilderSupport.*;
+
+interface RestrictionPredicate {
+
+    boolean accepts(UseRestriction r);
+}
+
 public class TextTranslationServiceImpl implements TextTranslationService {
 
     private OntologyTermSearchAPI api;
-    private OntModelCache ontModelCache = OntModelCache.INSTANCE;
 
     // This is a cache, used to associate a type (element of
     // the set { "disease", "organization", "commercial-status" }) to each named class.
@@ -26,15 +32,14 @@ public class TextTranslationServiceImpl implements TextTranslationService {
     // to cache it here in this class and not use it again if we don't need to.
     private Map<String, String> namedClassTypes;
     private OntModel model;
-    private StoreOntologyService storeOntologyService;
 
 
     @Inject
     public TextTranslationServiceImpl(StoreOntologyService storeOntologyService) {
-        this.storeOntologyService = storeOntologyService;
         this.namedClassTypes = new ConcurrentHashMap<>();
         try {
             Collection<URL> urls = storeOntologyService.retrieveOntologyURLs();
+            OntModelCache ontModelCache = OntModelCache.INSTANCE;
             model = ontModelCache.getOntModel(urls);
         } catch (Exception e) {
             throw new RuntimeException(e);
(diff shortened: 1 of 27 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -255,49 +255,38 @@ private String findNamedClassType(Named n) {
...
-        OntClass dataset_usage = model.getOntClass("http://www.broadinstitute.org/ontologies/DUOS/dataset_usage");
-        if (cls.hasSuperClass(dataset_usage)) {
+        OntClass datasetUsage = model.getOntClass(DATASET_USAGE);
+        OntClass duoPrimary = model.getOntClass(DUO_PRIMARY_CATEGORY);
+        if (cls.hasSuperClass(datasetUsage) || cls.hasSuperClass(duoPrimary)) {
             return "dataset_usage";
         }
-        return null;
+
+        return "other";
```
aednichols, line 275: The switch from `null` seems strange to me... is it to avoid a NPE somewhere? I worry that some person or code might think that `http://www.broadinstitute.org/ontologies/DUOS/other` is a real thing.

aednichols (approved):

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: That is correct, there is the case where something could be null. In this case, we're not actually constructing a new term here, we're using the string to make a categorical decision about what kind of translation to make, disease, population, etc.

davidangb (approved):
:+1:

PR #86, opened 2017-11-15
GAWB-2836: change search endpoint to be an exact-match by id

davidangb:
I'm not certain this is the right approach, as is much stricter than the previous behavior. This PR requires the input to the search endpoint to be the full http://purl.obolibrary.org/obo/... url, and it only performs exact matches.

Previous to this PR:
* the search behavior would accept part of the full url, such as "DOID_162"
* the search behavior would also return false positives for an unknown DOID url, such as "http://purl.obolibrary.org/obo/DOID_50847"

---

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Make sure Swagger is updated if API changes
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [ ] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Product Owner** sign off
- [ ] **Assign to submitter** to finalize
- [ ] **Submitter**: Verify all tests go green, including CI tests

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/service/ElasticSearchAutocompleteAPI.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/service/ElasticSearchAutocompleteAPI.java
@@ -9,6 +9,7 @@ import org.broadinstitute.dsde.consent.ontology.configurations.ElasticSearchConf
 import org.broadinstitute.dsde.consent.ontology.resources.model.TermParent;
 import org.broadinstitute.dsde.consent.ontology.resources.model.TermResource;
 import org.elasticsearch.client.Response;
+import org.elasticsearch.client.ResponseException;
 import org.elasticsearch.client.RestClient;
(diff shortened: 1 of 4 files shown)


rushtong (approved):
👍 I like this approach. Compliance uses full class ids in url form, and we should be enforcing this in Workbench.

PR #87, opened 2017-11-22
GAWB-2733: lib upgrades

davidangb:
Upgrading some dependencies due to proactive warnings from SourceClear.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
         <lucene.version>4.9.0</lucene.version>
         <dropwizard.version>1.0.0</dropwizard.version>
         <metrics.version>3.1.0</metrics.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -598,6 +610,10 @@
                     <groupId>msv</groupId>
                     <artifactId>xsdlib</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>xerces</groupId>
+                    <artifactId>xercesImpl</artifactId>
+                </exclusion>
```
davidangb, line 616: this exclusion isn't truly necessary, but we are explicitly pulling in xerces (line 562), so it felt right to exclude it here.

rushtong (approved):
👍 LGTM, tested all apis locally.

PR #88, opened 2017-12-15
GAWB-3007: read imported owl classes so the cache can be complete.

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-3007

## Notes
There are two ways that we currently look up ontology term ids. We have them indexed in ES, so there is a component that looks up ids there. Another, older way, is through the modified class here where we keep a cache of terms that are read on startup. This mechanism is far from ideal as it is not updated when files are updated. 

The change here is to make sure that, when using the cache lookup mechanism, we pull in imported ontology terms from any direct imports. This was the source of the failure. DUO terms live in an imported ontology, so were not placed in the cache and any restrictions that used them were counted as invalid due to "Term not found..." errors. 

See https://github.com/DataBiosphere/consent/blob/55907dd5291bc6f04ee6937f995d20db6f09c33e/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexerUtils.java#L96 for how we handled this in Consent when indexing files to Elastic Search.

The better (and a little bigger) change is to get rid of this mechanism altogether. There are currently two other services besides the Validation service that use it, but one of them is no longer in use and we have a story to update them: https://broadinstitute.atlassian.net/browse/GAWB-2590. Migrating that service to use the ES API will be the same category of work to migrate the Validate service as well.

---
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: JIRA ticket checks:
- [x] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [x] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [x] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [x] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [x] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Assign to submitter** to finalize
- [x] **Submitter**: Verify all tests go green, including CI tests

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/api/LuceneOntologyTermSearchAPI.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/api/LuceneOntologyTermSearchAPI.java
@@ -24,10 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/datause/api/LuceneOntologyTermSearchAPI.java
```diff
@@ -79,7 +78,13 @@ private void initAPI() throws OWLOntologyCreationException, IOException {
                     OWLAnnotationProperty def = annotationProperties.get(FIELD_DEFINITION_CLASS);
                     OWLAnnotationProperty deprecated = annotationProperties.get("deprecated");
 
-                    ontology.classesInSignature().forEach(owlClass -> {
+                    Set<OWLClass> owlClasses = ontology.classesInSignature().collect(Collectors.toSet());
+                    owlClasses.addAll(ontology.
+                            directImports().
+                            flatMap(HasClassesInSignature::classesInSignature).
+                            collect(Collectors.toSet()));
```
rushtong, line 85: This is the solution, pulling all direct imports into the cache.

Everything else is mostly just formatting changes.

davidangb (approved):
This looks right to me from the code level, and matches the errors I was seeing in practice (DUO nodes not found). I didn't test locally. :+1:

aednichols (approved):
👍

PR #97, opened 2018-04-03
GAWB-3210: Check for sub-conditions prior to defining GRU case. 

rushtong:
## Addresses
Jira: https://broadinstitute.atlassian.net/browse/GAWB-3210
See Also: https://github.com/broadinstitute/orsp/pull/342

## Changes
When constructing a structured use restriction from a data use object, check for sub-conditions in addition to the GRU case. It is now a valid case that GRU can be selected in addition to other fields, such as commercial use, etc. In the case where there are certain sub-conditions selected, we should not build a `{"type":"everything"}` SDUL and instead use the sub-conditions to generate the SDUL. 

---

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Make sure Swagger is updated if API changes
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [ ] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Product Owner** sign off
- [ ] **Assign to submitter** to finalize
- [ ] **Submitter**: Verify all tests go green, including CI tests

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilder.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilder.java
@@ -10,16 +10,41 @@ import static org.broadinstitute.dsde.consent.ontology.datause.builder.UseRestri
 
 /**
  * Apply consent-specific business rules when generating use restrictions
+ *
+ * There are several consent-related Data Use conditions, but the only ones
+ * that translate to a Structured Use Restriction are:
+ *  - General Use
+ *  - Disease Restrictions
+ *  - Population Restrictions (requires manual review)
+ *  - Commercial Use
+ *  - Gender
+ *  - Pediatric
+ *  - Methods Research
+ *  - Control Set
+ *
+ *  Other consent-related conditions include, but are currently not translated into an SDUR:
+ *  - HMB - Health/Medical/Biomedical Research
+ *  - POA - Population Origins Ancestry
+ *  - Population Structure
+ *  - Date Restriction (requires manual review)
+ *  - Aggregate Research (requires manual review)
+ *  - Recontacting Data Subjects
+ *  - Recontacting May
+ *  - Recontacting Must
+ *  - Genotypic Phenotypic data
+ *  - Cloud Storage
+ *  - Other and Other Restrictions (requires manual review)
+ *  - Ethics Approval Required
+ *  - Geographical Restrictions
+ *
+ *  Future expansions to DUOS could make use of these fields
+ *
  */
 public class ConsentRestrictionBuilder implements UseRestrictionBuilder {
 
     public UseRestriction buildUseRestriction(DataUse dataUse) {
         List<UseRestriction> categoryRestrictions = new ArrayList<>();
         UseRestriction restriction;
-
-        if (isPresent(dataUse.getGeneralUse()) && dataUse.getGeneralUse()) {
-            return new Everything();
-        }
 
         if (!dataUse.getDiseaseRestrictions().isEmpty()) {
             categoryRestrictions.add(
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilder.java
```diff
@@ -86,4 +126,15 @@ else if (dataUse.getGender().equalsIgnoreCase("female")) {
         return restriction;
     }
 
+    private Boolean isMethodsResearchProhibited(DataUse dataUse) {
+        if (isPresent(dataUse.getMethodsResearch()))
```
ahaessly: just want to double check that an empty value should default to false here, but above (line 108) an empty value defaults to true.

rushtong: Good question, and yes, I needed to triple check just to be sure this is correct. This is the code in ORSP that sets the value:
```
if (dataUseRestriction.methodsResearchExcluded) {
     dataUseDTO.methodsResearch = true
}
```
So ... 
* `methodsResearch:true` means that it is there is a restriction.
* `methodsResearch:false` means that it is there is no restriction.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilderTest.java
```diff
@@ -11,73 +12,155 @@
...
+     * sub-conditions exist that would render it as a non-GRU use restriction
+     */
+    @Test
+    public void testGeneralUseMixedCase1() {
+        DataUse dataUse = new DataUseBuilder().
+                setGeneralUse(true).
+                setDiseaseRestrictions(Collections.singletonList(ConsentUseCases.CANCER)).
+                build();
+        UseRestriction restriction = restrictionBuilder.buildUseRestriction(dataUse);
+        Assert.assertNotEquals(restriction, everything);
```
breilly2: I'm bothered by `assertNotEquals`. I've seen tests succeed with obviously wrong behavior because they used `assertNotEquals`. Can you instead check for what the result should be instead of what it shouldn't be?

rushtong: We have lots of tests that already test for what a restriction should be. This test is specifically for what it should not be.

breilly2: If there are tests for what a result should be then a separate test for what it should not be is redundant. Since these were the added tests (the rest of the changes in this file look like refactoring), I figured this was important coverage. Shouldn't there be other tests that have to change their expectations if the behavior is changing?

I'll also point out that I'm a bit out of my domain here so the coverage may be there but is not obvious to me.

rushtong: The original code made the assumption that if a `generalUse` flag was present, then the restriction would *always* be of that type. The requirements have changed such that we need to ignore that flag when other sub conditions exist. 

The existing tests all worked for the expected cases, and we never needed to check for this new case. They all still work even with the new logic flow, which is the most critical point.

These new tests are only for the singular purpose of the new case. They can certainly be considered redundant, but I'd rather have them and make sure the new assumptions still hold true rather than not having them at all.

rushtong: One more point to help clarify the point of our current tests. We have a extensive set of use cases that require specific outcomes for various combinations of choices. Those are each carefully constructed and tested for such that we get the right results when trying to match a research purpose and a consent. The test cases all map to something very specific. 

For the new test cases, it really doesn't matter what the end result of the restriction ends up being - we have no requirements for what a `generalUse` + `commercialUse` restriction needs to looks like. The only requirement we have is that it not be of `{"type":"everything"}`

rushtong: @breilly2 - I've thought about this some more and I think I can address your questions. I will take a stab at something more specific for these tests.

breilly2 (approved):
👍 Thanks for making changes to the tests. I like that much better!

ahaessly (approved):
👍

PR #104, opened 2018-11-11
[BTRX-102][risk=no] Update ontology's swagger implementation

rushtong:
## Addresses 
https://broadinstitute.atlassian.net/browse/BTRX-102


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
```diff
@@ -0,0 +1,87 @@
...
+            if (path.endsWith("png") || path.endsWith("gif")) {
+                byte[] content = FileUtils.readAllBytesFromResource(swaggerResource + path);
+                if (content != null) {
+                    response = Response.ok().entity(content).type(mediaType).build();
+                } else {
+                    response = Response.status(Response.Status.NOT_FOUND).build();
+                }
+            } else {
+                String content = FileUtils.readAllTextFromResource(swaggerResource + path);
+                if (content != null) {
```
vvicario: I think we can use StringUtils.isNotEmpty to verify if content is null or empty

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+
+@Path("/swagger")
+public class SwaggerResource {
+
+    private final static String swaggerResource = "META-INF/resources/webjars/swagger-ui/2.2.8/";
```
vvicario, line 17: what about if we move this path to config file, because if we update swagger version we will need to update this class

vvicario (approved):
Hi @rushtong code looks great I made a suggestion for swagger path.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
rushtong: I like that idea, but would like to hold off on that until some other PRs are merged. Will follow up with a story for it.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
rushtong: We can, thank you!

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
rushtong: Also, would like to tackle this for consent too.

PR #106, opened 2018-11-11
[BTRX-419][risk=no] Align templates with FC

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/BTRX-419
See https://github.com/broadinstitute/firecloud-orchestration for what we are aligning to.


LeoForconesi (approved):
Thanks for these changes! I think they are more practical to follow in the process.

PR #111, opened 2018-12-01
General Refactoring and Cleanup

rushtong:
Straight refactoring with no functional changes:
* Class renaming: favor `Service` over `API` ... I never understood why we named things this way.
* Lots of minor test fixes to clean up syntax warnings
* Update dropwizard, guice to the latest and remove unused injection lib.


pom.xml
```diff
@@ -482,67 +482,6 @@
             <version>4.11</version>
         </dependency>
 
-        <dependency>
-            <groupId>com.hubspot.dropwizard</groupId>
-            <artifactId>dropwizard-guice</artifactId>
```
rushtong, line 487: Turns out we are not actually using this library. It is also deprecated.

dags (approved):

PR #113, opened 2018-12-04
Handle not found and bad requests better

rushtong:


rushtong:
Got a 👍 from Walter in Slack.

PR #116, opened 2018-12-05
Exclude oauth block in swagger page

rushtong:
The ontology swagger page has a benign js error due to the lack of any auth security schemes. This PR removes the offending JS from the page since it is unnecessary.


LeoForconesi (approved):
:+1:

PR #124, opened 2019-01-23
[BTRX-452][risk=no] Guice Refactoring

rushtong:
Base refactoring PR to address https://broadinstitute.atlassian.net/browse/BTRX-452

See also:
* https://github.com/DataBiosphere/consent-ontology/pull/127
* https://github.com/DataBiosphere/consent-ontology/pull/128
* https://github.com/DataBiosphere/consent-ontology/pull/129


LeoForconesi (approved):
:+1:

PR #125, opened 2019-01-23
[BTRX-452][risk=no] minor refactoring

rushtong:
Small refactoring PR in my effort to make some progress on https://broadinstitute.atlassian.net/browse/BTRX-452

blocked on https://github.com/DataBiosphere/consent-ontology/pull/124

The goal with this set of stacked PRs is to clean up how we're using guice now so I can have a cleaner model to follow when updating consent. Consent will have some of the same problems we had here, but worse.
See also:
* https://github.com/DataBiosphere/consent-ontology/pull/124
* https://github.com/DataBiosphere/consent-ontology/pull/126
* https://github.com/DataBiosphere/consent-ontology/pull/127

PR #127, opened 2019-01-23
[BTRX-452][risk=no] Status refactoring

rushtong:
Small refactoring PR in my effort to make some progress on https://broadinstitute.atlassian.net/browse/BTRX-452

blocked on https://github.com/DataBiosphere/consent-ontology/pull/126

See also:
* https://github.com/DataBiosphere/consent-ontology/pull/124

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/StatusResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/StatusResource.java
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.consent.ontology.resources;
 
 import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheckRegistry;
+import com.google.inject.Inject;
 import org.apache.log4j.Logger;
 
 import javax.ws.rs.GET;
@@ -19,6 +20,7 @@ public class StatusResource {
 
     private HealthCheckRegistry healthChecks;
 
+    @Inject
     public StatusResource(HealthCheckRegistry healthChecks) {
         this.healthChecks = healthChecks;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/StatusResource.java
```diff
@@ -2,6 +2,7 @@
 
 import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheckRegistry;
+import com.google.inject.Inject;
```
tadeork, line 5: Just a question to understand. Are we using dropwizard-guicer or google-guice? Or is just that dropwizard-guicer implements this last one?

tadeork:
Looks good.

tadeork (approved):
Looks good

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/StatusResource.java
rushtong: We are just using google-guice. At one point, we were using the dropwizard plugin, but somehow we never really implemented it. I'm fuzzy on the history here. The plugin does provide some more shortcuts, but I don't really think they are worth it and add a layer of abstraction over what Google does, which is nice in and of itself. 

My goal with this story (and set of PRs) is to clean up how we're using guice now so I can have a cleaner model to follow when updating consent. Consent will have some of the same problems we had here, but worse :-(

rushtong: I'll add this summary to the PRs for better understanding.

rushtong: Done, thanks.

PR #128, opened 2019-01-23
[BTRX-452][risk=no] Use constructor injection

rushtong:
Small refactoring PR in my effort to make some progress on https://broadinstitute.atlassian.net/browse/BTRX-452

Makes no functional changes, changing to constructor injection

See also:
* https://github.com/DataBiosphere/consent-ontology/pull/124
* https://github.com/DataBiosphere/consent-ontology/pull/127
* https://github.com/DataBiosphere/consent-ontology/pull/129


tadeork (approved):
+1

PR #130, opened 2019-01-23
[BTRX-452][risk=no] More guice injection refactoring

rushtong:
More refactoring to address https://broadinstitute.atlassian.net/browse/BTRX-452

No functional changes, just moving to constructor-based injection.

Note that I am not worried about the loss in code coverage.

### DIFF
--- a/DEVNOTES.md
+++ b/DEVNOTES.md
@@ -90,4 +90,3 @@ elasticSearch:
 ```
 
 Consent will now point to a local ES instance. 
-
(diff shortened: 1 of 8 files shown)


tadeork (approved):
:+1:

rushtong:
Thank you @tadeork. I used this branch to test different coveralls settings as well. I like the coverage check merely to look at where the hotspots are for easy wins with tests and in general, I don't care what the overall percentage is. I would also like to know if we have a strong drop in coverage. To that end, I set the minimum coverage to 0% and the failure level to 10%. Hopefully that tones down the false alerts in coverage on PRs.

PR #135, opened 2019-04-01
Turn down mock server logging

rushtong:
Turn off mock server logging so tests can run more legibly.


vvicario (approved):
:+1:

rushtong:
Thank you @vvicario !

PR #136, opened 2019-04-02
Standardize Logging

rushtong:
No Ticket. Standardize all logging across the app.

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.consent.ontology.resources;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 import org.broadinstitute.dsde.consent.ontology.Utils;
 import org.parboiled.common.FileUtils;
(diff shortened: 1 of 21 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
```diff
@@ -38,11 +41,11 @@ private String getSwaggerResource() {
                 if (StringUtils.isNotEmpty(p.getProperty("swagger.ui.path"))) {
                     swaggerResource = p.getProperty("swagger.ui.path");
                 } else {
-                    logger.warning("swagger.ui.path is not configured correctly");
+                    log.log(Level.WARN, "swagger.ui.path is not configured correctly");
```
tadeork: Why here is ``log.log(Level.Warn,.. `` but in StatusResource.java is ``log.warn(...``? Is the same?

tadeork (approved):
It's OK. I asked one question just to understand.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/SwaggerResource.java
rushtong: Good question ... they should be the same. For some reason, the logger here didn't have a warn method ... something isn't quite right there. Will look at it.

rushtong: OK, my bad - there is a `warn`. For some reason, my IDE wasn't compiling correctly last night. Will fix.

tadeork: Great then. That was the only thing that I came across and call my attention.

rushtong: Done!

PR #137, opened 2019-04-03
[DUOS-231] Algorithm Documentation

rushtong:
## Addresses 
See https://broadinstitute.atlassian.net/browse/DUOS-231

## Changes
* Lots of swagger yaml updates to fix some consistency issues
* Add a `match/v1` endpoint that calls the original match method
* Move the `match/dataUse` matching endpoint (currently not in use) to `match/v2`
* Docs for the two algorithms we have now. For V2, there is a [google doc here](https://docs.google.com/document/d/1P70Gt5wCru0YzvJWNru9Nt4tCADhEsBWpKPp7qs_n2M) that describes the business rules developed for FireCloud implementation. Our implementation of the rule set is a bit more expansive than what's covered there. We handle gender and pediatric conditions where FC does not. 
* Docs for use restriction grammar. This is [originally here](https://broadinstitute.atlassian.net/wiki/spaces/DSDEV/pages/8323107/Consent+and+Use+Restriction+Grammar) but that version is very out of date. I have made some small edits and pruned it a bit.
* Consolidated docs to a docs directory.

The algorithm readme looks something like this PDF export: [Algorithms.pdf](https://github.com/DataBiosphere/consent-ontology/files/3048007/Algorithms.pdf)

The use restriction grammar readme looks something like this PDF export: [UseRestrictionGrammar.pdf](https://github.com/DataBiosphere/consent-ontology/files/3039648/UseRestrictionGrammar.pdf)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,16 @@ A semi-automated management service for compliant secondary use of human genomic
 There are restrictions on researching human genomics data. For example: “Data can only be used for breast cancer research with non-commercial purpose”.
 The Data Use Oversight system ensures that researchers using genomics data honor these restrictions.
 
-See [consent-ontology/DEVNOTES.md](DEVNOTES.md) to run locally.
+## Matching Algorithms
+
+See [Algorithm](docs/Algorithms.md) and [Use Restriction Grammar](docs/UseRestrictionGrammar.md) for detailed information on how the system 
+peforms data use matching between consented datasets and research purposes.
+
+## Contributing
+
+See [Contributing](docs/CONTRIBUTING.md) for information on how to 
+contribute to this project. 
+
+## Local Development
+See [DEVNOTES.md](docs/DEVNOTES.md) to run locally.
+
--- a/docs/Algorithms.md
+++ b/docs/Algorithms.md
@@ -1,0 +1,60 @@
+# Matching Algorithm
+
+See the [Swagger](https://consent-ontology.dsde-prod.broadinstitute.org/#/) page for detailed usage
+of the Matching APIs. This document describes the different approaches to matching a Research Purpose
+and a Consent on a dataset. Both Purpose and Consent are structured objects which allows for computational
+matching at scale. 
+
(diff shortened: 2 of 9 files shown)


kreinold (approved):
1. It's a little confusing that you've switched from DUO terminology of "data use" to "research purpose" in the description of the matching algorithm.  Do they mean exactly the same thing?
2. Rather than keeping a running history of the version descriptions here, it would be good to replace the previous versions when revising the file.  However, I realize that you are adding a description for Version 1 where there was none before, so this is fine, but in future....

rushtong:
Thank you @kreinold! I really appreciate the effort to look this over - it's non-trivial. 

Re: 1: In our world of algorithmic matching, `data use` is structured thing that can apply to either a research purpose or a consented dataset. In this codebase, we have two distinct objects that represent `data use`. We have the original version which is a [`UseRestriction`](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/models/UseRestriction.java) and we have a newer, more simplified json object, [`DataUse`](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/model/DataUse.java). In essence, `UseRestriction` and `DataUse` are conceptually the same thing, but just different implementations of the idea of a set of restrictions that can apply to either a research purpose or a consented dataset.

Re: 2: The point here is to document what we currently support. Since we support both versions, we need to have documentation that describes them. The history of this is that for the past 4 years or so, our Data Access Committee has been working with the results of `V1`. Future work in this area will allow us to show the results of the matching from not only the `V1`, but from `V2`, and perhaps someday, `V3`. This version comparison will allow us to show the DAC that our algorithm has evolved and is, in theory, getting us closer and closer to 100% agreement with the DAC's decisions.

rushtong:
Got a verbal 👍 from Jonathan in standup today.

PR #139, opened 2019-05-28
[DUOS-367][riskno] fix ncu/npu translation

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/DUOS-367

Update NCU/NPU logic when translating data use json to structured use restriction.

### DIFF
--- a/src/test/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilderTest.java
+++ b/src/test/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilderTest.java
@@ -9,6 +9,8 @@ import org.junit.Test;
 
 import java.util.Collections;
 
+import static org.broadinstitute.dsde.consent.ontology.datause.builder.UseRestrictionBuilderSupport.NON_PROFIT;
+
 public class ConsentRestrictionBuilderTest {
 
     private UseRestrictionBuilder restrictionBuilder = new ConsentRestrictionBuilder();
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilderTest.java
```diff
@@ -219,4 +221,24 @@ public void testCSdulUC3() {
...
+        DataUse dataUse = new DataUseBuilder().
+                setCommercialUse(true).
+                build();
+        UseRestriction restriction = restrictionBuilder.buildUseRestriction(dataUse);
+        Assert.assertTrue(restriction.toString().contains(NON_PROFIT));
+        Assert.assertTrue(restriction.toString().contains("\"type\":\"not\""));
+    }
+
+    @Test
+    public void testCommercialUseExcluded() {
```
danxmoran, line 235: Should there be a test for `getCommercialUse` returning `null`, since the main code tries to handle it?

danxmoran (approved):
One suggestion, but looks good as far as I can tell 😄

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/builder/ConsentRestrictionBuilderTest.java
rushtong: Nice catch, and yes, there should

PR #143, opened 2019-06-06
[DUOS-331][risk=no] Add failure reasons to the V2 match logic

rushtong:
## Addresses
See https://broadinstitute.atlassian.net/browse/DUOS-331

## Changes
Mostly a refactor of the individual match cases so we can add in failure strings to give the recipient more information about why the algorithm failed a particular purpose <-> consent match. 

## TODO
* Still need to run language by PO
* Increase test cases to check for expected errors

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCases.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCases.java
@@ -23,6 +23,18 @@ class DataUseMatchCases {
     private static final String na = "N/A";
     private static final String male = "Male";
     private static final String female = "Female";
+    private static final String HMB_F1 = "The GRU Research Purpose does not match the HMB data use limitations.";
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCases.java
```diff
@@ -107,22 +118,27 @@ static boolean matchDiseases(DataUse purpose, DataUse dataset, Map<String, List<
...
         }
 
+        List<String> failures = new ArrayList<>();
         boolean datasetNMDS = getNullableOrFalse(dataset.getMethodsResearch());
         if (!datasetNMDS) {
-            return true;
+            return ImmutablePair.of(true, Collections.emptyList());
+        } else {
+            failures.add("[NMDS] Dataset is restricted to no method development research");
         }
```
rushtong, line 146: Should we return this failure right away?

vvicario (approved):
Nice  job!

tadeork (approved):
I think this is a lot of work. So the decrease of the coverage is reasonable.

PR #199, opened 2020-01-27
[DUOS-498][risk=no] Circleci configs

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-498

## Changes
* Use circle instead of travis.
* Cobertura was complaining about parsing `<` in the test result files. Looks like the plugin hasn't been updated in quite some time. Ditch that for jacoco which seems a little more forgiving.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   jdk:
     docker:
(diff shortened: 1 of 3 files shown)


Marianie-Alphonse (approved):

rushtong:
Thank you @Marianie-Simeon !

PR #208, opened 2020-02-05
[DUOS=531][risk=no] Update to swagger 3

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-531

## Changes
* Upgrade the swagger ui version to the latest

### New UI
Screen Shot 2020-02-13 at 8 00 31 AM

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,7 +140,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.2.1</version>
+                <version>3.2.2</version>
                 <configuration>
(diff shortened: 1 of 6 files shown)


verovicario (approved):
:+1:

pom.xml
```diff
@@ -563,73 +563,73 @@
         <dependency>
             <groupId>org.eclipse.rdf4j</groupId>
             <artifactId>rdf4j-rio-binary</artifactId>
-            <version>${rdf4j-rio.version}</version>
+            <version>3.0.4</version>
```
verovicario: I think we can create a custom variable with 3.0.4 version number

rushtong: Oops, this is a bad merge - we already have that in dev. Will fix.

PR #225, opened 2020-02-13
[DUOS-566][risk=no] Handle missing ontology terms better

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-566

## Changes
* Catch errors with purpose terms before continuing on with other algorithm checks

### DIFF
--- a/src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherTest.java
+++ b/src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherTest.java
@@ -11,7 +11,6 @@ import org.junit.Test;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherTest.java
```diff
@@ -271,11 +270,7 @@ private void assertNegative(DataUse purpose, DataUse dataset) {
     private ImmutablePair<Boolean, List<String>> matchPurposeAndDataset(DataUse purpose, DataUse dataset) {
         DataUseMatcher matcher = new DataUseMatcher();
         matcher.setAutocompleteService(autocompleteService);
-        try {
```
Marianie-Alphonse, line 274: Why did you remove the try-catch error handling here? Does it get handled in another place?

rushtong: Yes, exactly, line 43 of `DataUseMatcher` is the method signature which no longer throws an exception. Inside the method, we handle the error case.

Marianie-Alphonse: sounds good!

Marianie-Alphonse (approved):

PR #304, opened 2020-08-27
[DUOS-740][risk=no] Update match v1 resource

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-740

## Changes
This bug is not reproducible locally, but it's the akka system that is shutting down the server on failure. This PR removes the unnecessary actor system and instead handles the async request directly. This might uncover other issues that the akka shutdown was masking and we'll deal with them as they come up.


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -29,23 +14,28 @@
...
-
-    private static final ActorSystem actorSystem = ActorSystem.create("actorSystem");
-    private static final ActorRef matchWorkerActor = actorSystem.actorOf(Props.create(MatchWorkerActor.class), "MatchWorkerActor");
+    private final StoreOntologyService storeOntologyService;
+    private final DataUseMatcher dataUseMatcher;
+    private final OntModelCache ontModelCache = OntModelCache.INSTANCE;
 
-    MatchResource() {}
+    @Inject
+    MatchResource(DataUseMatcher dataUseMatcher, StoreOntologyService storeOntologyService) {
```
rushtong, line 35: Moved to constructor injection per https://github.com/google/guice/wiki/MinimizeMutability

lindojong (approved):
Changes make sense to me. I especially like how the code for matchV1 was cleaned up!

PR #365, opened 2021-01-06
[DUOS-908][risk=no] Use DSP blessed image

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-908

---- 

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/DataBiosphere/consent-ontology/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/DataBiosphere/consent-ontology/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing
- [ ] I've updated Swagger to reflect any API changes

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,3 @@
-FROM adoptopenjdk:8-hotspot
-
-# Standard apt-get cleanup.
-RUN apt-get -yq autoremove && \
-    apt-get -yq clean && \
-    rm -rf /var/lib/apt/lists/* && \
-    rm -rf /tmp/* && \
-    rm -rf /var/tmp/*
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:8-distroless
 
 COPY target/consent-ontology.jar /opt/consent-ontology.jar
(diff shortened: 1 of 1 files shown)


lindojong (commented):
Just curious since it seems they recommend distroless over alpine, would switching over require significant work on our part, or is there some other reason we should be sticking with alpine?

rushtong:
@lindojong 

> Just curious since it seems they recommend distroless over alpine, would switching over require significant work on our part, or is there some other reason we should be sticking with alpine?

For Ontology, the distroless works out of the box. For Consent, I ran into errors on startup. That could have just been my local set up, but I didn't want to spend too much time digging into it. I'm sure we could use distroless with some more investigation work, but alpine worked without any changes and is still a blessed image.

lindojong (approved):

PR #671, opened 2022-08-19
[DUOS-2028][risk=no] Implement GET (/translate/paragraph) for Data Use Ontology recommendations in Consent Ontology

gabelmattos:
----

- Verify all tests go green


src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,6 +89,13 @@ public String translateDataset(String dataUseString) throws IOException {
         return translate(dataUse, TranslateFor.DATASET);
     }
 
+    @Override
+    public String translateParagraph(String dataUseString) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
```
pshapiro4broad: `ObjectMapper` is thread safe so most applications only need to instantiate one copy. In an app that uses IOC (inversion of control) (that is, one that uses `Inject` or `Autowired`) the usual approach is to make it a singleton that can be injected, in case you want to configure it specially for you app.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,6 +89,13 @@ public String translateDataset(String dataUseString) throws IOException {
...
+        DataUse dataUse = mapper.readValue(dataUseString, DataUse.class);
```
pshapiro4broad: As a side node, this code is using both `Gson` and `ObjectMapper`. These are two different libraries that do (pretty much) the same thing, it would be better to only use one or the other.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
```diff
@@ -3,15 +3,16 @@
...
 import org.broadinstitute.dsde.consent.ontology.datause.services.TextTranslationService;
+import org.broadinstitute.dsde.consent.ontology.translate.DTO.RecommendationDTO;
+import org.broadinstitute.dsde.consent.ontology.translate.service.Translate;
 import org.slf4j.Logger;
 
-import javax.ws.rs.Consumes;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.QueryParam;
+import javax.ws.rs.*;
```
pshapiro4broad: Not sure what the conventions are in this project, but I usually avoid wildcard imports in Java since it makes it harder to tell where a name comes from when reading the code.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
```diff
@@ -53,6 +54,20 @@ public Response translate(@QueryParam("for") String forParam, String restriction
         }
     }
 
+    @Path("paragraph")
+    @GET
+    public Response translateParagraph(@QueryParam("paragraph") String paragraph) throws Exception {
+        if (paragraph == null || paragraph.isEmpty()) {
```
pshapiro4broad: It's a shame you need to null check here; I know swagger v3 has support for required but I don't think javax.ws does. Using `@Nonnull` on the parameter might work to generate an exception at runtime.

If you do this a lot, you can use apache commons lang3 StringUtils. https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
```diff
@@ -53,6 +54,20 @@ public Response translate(@QueryParam("for") String forParam, String restriction
...
+            return Response.status(Response.Status.BAD_REQUEST).entity("Missing parameter: paragraph").build();
+        }
+
+        Translate translate = new Translate("search-terms.json");
```
pshapiro4broad: `Translate` is in a `service` package but doesn't look like a service class. If you always instantiate it using `"search-terms.json"` then it would make sense to make it injectable and inject it into this class, and then it would be a service class.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
```diff
@@ -53,6 +54,20 @@ public Response translate(@QueryParam("for") String forParam, String restriction
...
+
+        HashMap<String, RecommendationDTO> recommendations = translate.paragraph(paragraph);
+
+        return Response.ok().entity(recommendations).build();
```
pshapiro4broad: Minor improvement
```suggestion
        return Response.ok(recommendations).build();
```

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/DTO/RecommendationDTO.java
```diff
@@ -0,0 +1,24 @@
+package org.broadinstitute.dsde.consent.ontology.translate.DTO;
+
+public class RecommendationDTO {
```
pshapiro4broad: The java naming conventions most DSP projects follow (based on the google java conventions) require that package names are all lower case and that initialisms like `DTO` only have the first letter capitalized. 
```suggestion
package org.broadinstitute.dsde.consent.ontology.translate.dto;

public class RecommendationDto {
```

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/DTO/RecommendationDTO.java
```diff
@@ -0,0 +1,24 @@
...
+    private String title;
```
pshapiro4broad: When possible I would make fields final. In this case you can remove the setters and use a constructor and then the fields could be final.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/SearchUtils.java
```diff
@@ -0,0 +1,18 @@
...
+public class SearchUtils {
+    public static boolean searchForKeyword(final String keyword, final String targetText) {
+        if (Utils.isNullOrEmpty(keyword) || Utils.isNullOrEmpty(targetText)) {
+            return false;
+        }
+
+        return Pattern
+                .compile(keyword, Pattern.CASE_INSENSITIVE)
+                .matcher(targetText)
+                .find();
```
pshapiro4broad: Unless `keyword` contains a regular expression (or using `Pattern` is more performant), you can do a substring search here. Using apache commons StringUtils handles the null checks, unless you want to return `false` for a substring search for the empty string, which would normally return `true`.
```
    return StringUtils.containsIgnoreCase(targetText, keyword);
```

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/LoadUtils.java
```diff
@@ -0,0 +1,23 @@
...
+import com.github.jsonldjava.shaded.com.google.common.reflect.TypeToken;
+import com.google.gson.Gson;
+import org.broadinstitute.dsde.consent.ontology.translate.model.TermItem;
+import org.parboiled.common.FileUtils;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+public class LoadUtils {
+    public static List<TermItem> termsFromLocalResourcesJSON(@NotNull String jsonFilePath) throws Exception {
```
pshapiro4broad: I would leave this inside `Translate` until you have another use for it outside of that class. In general unless you know something will be useful as a separate "utility" method I don't think it's worth making a separate class for it.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/LoadUtils.java
```diff
@@ -0,0 +1,23 @@
...
+        try {
+            final String searchTerms = FileUtils.readAllTextFromResource(jsonFilePath);
+            return new Gson().fromJson(
+                    searchTerms,
+                    new TypeToken<List<TermItem>>() {}.getType()
+            );
```
pshapiro4broad: I would write this using `ObjectMapper.readValue()`. It has a form that accepts a `File` to read from.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/LoadUtils.java
```diff
@@ -0,0 +1,23 @@
...
+        } catch (Exception e) {
+            throw new Exception("Error loading terms from json file: " + jsonFilePath, e);
```
pshapiro4broad: I don't know about this code base, but normally I wouldn't catch or throw Exception (or declare `throws Exception`) outside of test code. For `catch` I'd catch a more specific type. For `throw` I'd throw an exception type defined in my code. And unless the exception is expected to be caught by the caller, I'd use a `RuntimeException` subclass so it doesn't need to be declared as a checked exception. (Checked exceptions are fine too if that's what a code base is using.)

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/SearchUtils.java
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsde.consent.ontology.translate.utils;
+
+import org.broadinstitute.dsde.consent.ontology.Utils;
+
+import java.util.regex.Pattern;
+
+public class SearchUtils {
+    public static boolean searchForKeyword(final String keyword, final String targetText) {
```
pshapiro4broad: I would put this inside `Translate` too.

src/test/java/org/broadinstitute/dsde/consent/ontology/service/MockTranslationService.java
```diff
@@ -15,6 +15,11 @@ public String translateDataset(String dataUse) {
         return "translated dataset";
     }
 
+    @Override
+    public String translateParagraph(String paragraph) {
+        return "{\"http://purl.obolibrary.org/obo/DUO_0000007\":{\"title\":\"Disease Specific Research\",\"category\":\"Data Use Permission\"}}";
```
pshapiro4broad: It would probably be better to make the tests independent of the actual `search-terms.json`. You'd have to make `Translate` injectable and make the json it reads in a configuration object.

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResourceTest.java
```diff
@@ -39,6 +39,18 @@ public void testDatasetTranslate() throws IOException {
...
+    @Test
+    public void testParagraphTranslate() throws IOException {
+        String paragraph = "This could be a disease or anything else.";
+        Response response = gRule.client().target("/translate/paragraph")
+                .queryParam("paragraph", paragraph)
+                .request(MediaType.APPLICATION_JSON_TYPE)
+                .accept(MediaType.TEXT_PLAIN_TYPE)
+                .get();
+        assertEquals(200, response.getStatus());
+        assertEquals(mockService.translateParagraph(paragraph), response.readEntity(String.class));
```
pshapiro4broad: Note that this is testing two classes, the `TranslateResource` API and the `Translate` class. If `Translate` was injectable then you could test these classes independently.

src/main/java/org/broadinstitute/dsde/consent/ontology/Utils.java
```diff
@@ -9,4 +9,8 @@ public static Logger getLogger(Class clazz) {
         return LoggerFactory.getLogger(clazz);
     }
 
+    public static boolean isNullOrEmpty(final String s) {
```
pshapiro4broad: I think `StringUtils.isEmpty()` is the same as this

pshapiro4broad (commented):
Still looking but here are a few comments

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
gabelmattos: Great point.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
gabelmattos: I thought about that annotation, but I wasn't sure of the mechanics of how that annotation is in this context. I guess I can try it out.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/SearchUtils.java
gabelmattos: That sounds preferable.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/SearchUtils.java
gabelmattos: My thoughts were, "can I abstract this to use elsewhere?"
And I believe it may be applicable in other future things, or I may want to expand how the search functionality works. 

But I see your point; translating is the better option if this is not usable elsewhere.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: As a side note, we've been trying to migrate away from using Jackson libraries, in general. I much prefer Gson for this kind of job.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
rushtong: We don't use * imports. You can configure intellij to handle this with a very large number, like 9999:

Screen Shot 2022-08-22 at 8 26 19 AM

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
rushtong: +1 to `StringUtils` .

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/utils/SearchUtils.java
pshapiro4broad: My rule of thumb is, whenever I say "I may need this in the future", I know I probably shouldn't be doing it now. Until that future day comes, it's better to write the code as it's needed today. The internet name for this is YAGNI.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
pshapiro4broad: When you have time I'd be interested to know why you prefer gson over Jackson. The main reason I use Jackson is because most DSP java apps use Spring, and Spring uses Jackson, and I'd rather use the same library as spring (since my app is already importing that library) unless I have a good reason not to. And so far I haven't run into any issues or shortcomings using Jackson.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
pshapiro4broad: In our repo I use an `.editorconfig` file to automatically set this in IntelliJ. https://github.com/DataBiosphere/terra-data-catalog/blob/main/.editorconfig

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
```diff
@@ -0,0 +1,52 @@
...
+            for (String keyword : keywords) {
+                final boolean foundMatch = SearchUtils.searchForKeyword(keyword, paragraph);
+
+                if (foundMatch) {
+                    RecommendationDTO recommendation = new RecommendationDTO();
+                    recommendation.setTitle(title);
+                    recommendation.setCategory(category);
+                    if (!recommendations.containsKey(url)) {
+                        recommendations.put(url, recommendation);
+                    }
```
pshapiro4broad: This case can arise two ways. One is if a `term` with the same `url` appears multiple times. Is that an expected case from the JSON data used to generate the list of terms?

The other way is if you find more than one match. To handle that case, you can `break` out of the inner `for`. For example:
```suggestion
                    recommendations.put(url, recommendation);
                    break;
```

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public HashMap<String, RecommendationDTO> paragraph(final String paragraph) throws Exception {
+        HashMap<String, RecommendationDTO> recommendations = new HashMap<>();
+
+        if (Utils.isNullOrEmpty(paragraph) || Utils.isNullOrEmpty(jsonFilePath)) {
+            return recommendations;
+        }
+
+        List<TermItem> terms = LoadUtils.termsFromLocalResourcesJSON(jsonFilePath);
```
pshapiro4broad: It looks like this list of terms is constant over the lifetime of the app. In that case, if it's not too large, I would load the list of terms once during `Translate`'s construction. The overhead of loading a json structure isn't great but there's no benefit in incurring this cost every time an API call is made. In the case where the list of terms is very large, you'd probably need a better way to perform an indexed search over terms.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
```diff
@@ -0,0 +1,52 @@
...
+
+        for (TermItem term : terms) {
+            final String title = term.getTitle();
+            final String category = term.getCategory();
+            final String url = term.getUrl();
+            final String[] keywords = term.getKeywords();
+
+            for (String keyword : keywords) {
+                final boolean foundMatch = SearchUtils.searchForKeyword(keyword, paragraph);
```
pshapiro4broad: If the only way a term can be matched is via its keywords, a better way to represent this may be to make a associative array that maps every keyword to its `TermItem`s, maybe `Map<String, List<TermItem>>`. Then you could search over all words in a paragraph and for each word do a keyword lookup (which is O(1)) so the term search would be O(n) instead of O(n^2). You'd have to combine all found terms into a single list and then convert that into your DTO.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
gabelmattos: A recommendation shows up multiple times. If the phrase contains "gru general and general research" all in the same paragraph, then it'll recommend GRU 3 times.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
gabelmattos: Reverse index with keywords? Clever. I like that.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
gabelmattos: I'm unclear whether making it required in swagger extends to external calls. 
Does it?

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: I have two issues with Jackson. One is that it had been a large source of security vulnerabilities and required many frequent updates. That has subsided over the past year so is less of an issue today. The other one is the amount of boilerplate code necessary with model class annotations. I find Gson far more flexible from that perspective.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
rushtong: Making it required in swagger has no effect in the java code. You'd need to handle that non-nullability in the resource class too.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class Translate {
+
+    public HashMap<String, RecommendationDto> paragraph(final String paragraph) throws Exception {
+        HashMap<String, RecommendationDto> recommendations = new HashMap<>();
+
+        List<TermItem> terms = loadJSONFromResources("search-terms.json");
```
rushtong: Ontology has a pattern for reading resource files like this. We store them in a bucket and read from that. This lets us modify the file on the fly and bounce any affected pods if necessary. Admittedly, our GCS code is very old in this repo and needs to be completely re-written since Google has updated their java clients. Look for usages of `configurationFileName` which is essentially this file (configurable per environment):
```
{
  "https://storage.googleapis.com/broad-dsde-dev-consent/ontology/disease/diseases.owl": {
    "fileName": "diseases.owl",
    "prefix": "DOID",
    "ontologyType": "Disease"
  },
  "https://storage.googleapis.com/broad-dsde-dev-consent/ontology/organization/data-use.owl": {
    "fileName": "data-use.owl",
    "prefix": "DUOS",
    "ontologyType": "Organization"
  }
}
```
I would rather we maintain the GCS pattern as opposed to the current one in this PR.

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResourceTest.java
rushtong: I agree with Phil's suggestion here. We should test the service and the resource separately. There are some poor testing practices on display in this test class. Let's pull the service call out and put that in a service class test.

src/main/resources/assets/api-docs.yaml
```diff
@@ -116,6 +116,27 @@ paths:
...
+      summary: Translates Paragraph to ontology recommendations
+      description: The Translate service for paragraphs will render an ontology recommendation for a given paragraph.
+      produces:
+        - application/json
+      parameters:
+        - name: paragraph
+          in: query
+          description: Paragraph to translate
+          type: string
+          required: true
```
rushtong: Our services run behind an apache proxy which, I think, limits the url length to 8177 characters. I did some playing around with random strings and I think this could be hit without much difficulty. Maybe make this a POST that accepts `text/plain` to get past that?

rushtong: For example, this is what we get with a long request:
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>414 Request-URI Too Long</title>
</head><body>
<h1>Request-URI Too Long</h1>
<p>The requested URL's length exceeds the capacity
limit for this server.<br />
</p>
<hr>
<address>Apache Server at local.broadinstitute.org Port 443</address>
</body></html>
```

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
gabelmattos: I still need synonyms. I can't use the owl. 
I could use elasticsearch. But I need to add the synonyms.

src/main/resources/assets/api-docs.yaml
gabelmattos: Good call.

src/main/java/org/broadinstitute/dsde/consent/ontology/translate/service/Translate.java
rushtong: Maybe my suggestion wasn't clear ... I'm not suggesting that you use the actual file I quoted. I'm suggesting that we move `search-terms.json` to the same GCS bucket where we store `configuration.json` (which is here: https://console.cloud.google.com/storage/browser/broad-dsde-dev-consent/ontology). Then use the GCS service class to load this file up instead of reading it from `resources`.

gabelmattos: Ah, ok, that makes sense then.

pom.xml
```diff
@@ -740,6 +740,12 @@
             <artifactId>swagger-ui</artifactId>
             <version>${swagger.ui.version}</version>
         </dependency>
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter</artifactId>
+        <version>RELEASE</version>
+        <scope>test</scope>
+      </dependency>
```
rushtong: I think we should revert this file, changes from this PR shouldn't have these in it.

src/main/java/org/broadinstitute/dsde/consent/ontology/OntologyModule.java
```diff
@@ -39,7 +38,7 @@ protected void configure() {
     @Provides
     @Singleton
     public TextTranslationService providesTextTranslationService() {
-        return new TextTranslationServiceImpl(providesAutocomplete());
+        return new TextTranslationServiceImpl(providesAutocomplete(), providesGCSStore(), providesStore());
```
rushtong: We aren't using `providesStore()` in the impl class so we should remove it here.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -69,11 +76,14 @@ public class TextTranslationServiceImpl implements TextTranslationService {
...
     private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("MM/dd/yyyy");
 
-    private AutocompleteService autocompleteService;
+    private final AutocompleteService autocompleteService;
+
+  private final GCSStore gcsStore;
 
     @Inject
-    public TextTranslationServiceImpl(AutocompleteService autocompleteService) {
+    public TextTranslationServiceImpl(AutocompleteService autocompleteService, GCSStore gcsStore, StoreOntologyService storageService) {
```
rushtong: `storageService` is unused:
```suggestion
    public TextTranslationServiceImpl(AutocompleteService autocompleteService, GCSStore gcsStore) {
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +99,65 @@ public String translateDataset(String dataUseString) throws IOException {
...
+    }
+
     @Override
     public String translatePurpose(String dataUseString) throws IOException {
         ObjectMapper mapper = new ObjectMapper();
         DataUse dataUse = mapper.readValue(dataUseString, DataUse.class);
         return translate(dataUse, TranslateFor.PURPOSE);
     }
 
+    public HashMap<String, Recommendation> paragraph(final String paragraph) throws Exception {
```
rushtong: This can be private

src/main/resources/assets/api-docs.yaml
```diff
@@ -365,6 +387,12 @@ definitions:
     example:
       { "generalUse": false, "hmbResearch": true, "diseaseRestrictions": ["http://purl.obolibrary.org/obo/DOID_602"] }
 
+  DataUseRecommendation:
+    type: object
+    description: JSON model of a set of data use recommendations
+    example:
+      { "http://purl.obolibrary.org/obo/DUO_0000042": { title: "General Research Use", "category", "Data Use Permission" } }
+
```
rushtong: I think what you want here is this:
```suggestion
    example:
      { "http://purl.obolibrary.org/obo/DUO_0000042": {"title": "General Research Use", "category": "Data Use Permission" } }
```

This seems to correlate more closely with map of String -> Recommendation which is the response object.

rushtong: We don't have a real `DataUseRecommendation` class. Perhaps we can create one and use that as the actual response object in the resource class?

src/main/resources/assets/api-docs.yaml
```diff
@@ -391,6 +419,13 @@ definitions:
           Optional list of reasons why a match result may have failed. Only supported in
           in versions 2 and higher
 
+  TranslateParagraph:
+    type: object
+    properties:
+      paragraph:
+        description: The paragraph to translate
+        type: string
```
rushtong, line 426: This is a great place for an example. Also, if we had a class for this, we could deserialize the JSON with that instead of the more generic form of:
```
final Map<String, String> body = mapper.readValue(jsonString, new TypeReference<>() {});
```
... and then we would not have to assume that `String paragraph = body.get("paragraph");` would be a valid call, we could instead just say:
```
TranslateParagraph payload = new Gson().fromJson(jsonString, TranslateParagraph.class);
String paragraph = payload.getParagraph();
```

src/main/resources/search-terms.json
```diff
@@ -0,0 +1,80 @@
+[
```
rushtong: I think we need to remove this version of the file in `/src/main/resources` and keep the one in `/src/test/resources`

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -89,6 +112,19 @@ public void testTranslateDataset() throws IOException {
...
+      HttpResponse response = loadTermItemsResponse();
+      when(gcsStore.getStorageDocument(Mockito.anyString())).thenReturn(response);
+      String mockParagraph = "GRU General research for some test with disease. This is not for profit.";
+
+      HashMap<String, Recommendation> translation = service.translateParagraph(mockParagraph);
+      assertEquals(3, translation.size());
+      assertEquals("General Research Use", translation.get("http://purl.obolibrary.org/obo/DUO_0000042").title());
+      assertEquals("Disease Specific Research", translation.get("http://purl.obolibrary.org/obo/DUO_0000007").title());
+      assertEquals("Not for Profit Organization Use Only", translation.get("http://purl.obolibrary.org/obo/DUO_0000045").title());
+    }
```
rushtong: Great test. I think we could also use a service level test for some interesting cases:

- "Science"
- ""
- "for profit"
- "FOR PROFIT"

This is also a great fit for parameterized tests: https://www.baeldung.com/parameterized-tests-junit-5

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResourceTest.java
```diff
@@ -8,8 +8,12 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.MockitoAnnotations.openMocks;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.client.http.HttpStatusCodes;
 import com.google.gson.Gson;
 import javax.ws.rs.core.Response;
+
+import org.broadinstitute.dsde.consent.ontology.cloudstore.GCSStore;
```
rushtong: Looks like these are unused, BUT, we should add some new tests to handle the new resource method, `translateParagraph`.

rushtong (changes requested):
Nice work overall! See comments inline. Also, let me know if you would prefer this - I can submit a cleanup PR on this one which can address all of my comments. That could then be re-reviewed by the team.

src/main/java/org/broadinstitute/dsde/consent/ontology/model/TermItem.java
```diff
@@ -0,0 +1,41 @@
+package org.broadinstitute.dsde.consent.ontology.model;
+
+public class TermItem {
```
rushtong, line 3: I was able to make this work as a record class, but Gson doesn't work well with records yet (not sure exactly why, but I am looking into it), so you'd need to switch to ObjectMapper to turn the json into a list of terms. This is really just a note for future work ... Gson really should be working out of the box and it's problematic that it isn't here.

pom.xml
```diff
@@ -1,745 +1,745 @@
...
+  <name>Consent Management: Ontology Services</name>
+
+  <properties>
+    <dropwizard.version>2.1.1</dropwizard.version>
+    <java.version>17</java.version>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+    <jena.version>2.6.4</jena.version>
+    <jersey.version>2.19</jersey.version>
+    <mockserver.version>5.14.0</mockserver.version>
```
rushtong: It looks like this file has been reformatted. From the specific diffs, I think this file can be reverted - I don't see any new imports here.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,65 @@ public String translateDataset(String dataUseString) throws IOException {
...
+      return recommendations;
+    }
+
+    private static boolean searchForKeyword(final String keyword, final String targetText) {
+      return StringUtils.containsIgnoreCase(targetText, keyword);
+    }
+
+    List<TermItem> loadTermsFromGoogleStorage() throws Exception {
+      HttpResponse response = gcsStore.getStorageDocument("ontology/search-terms.json");
+      String terms = response.parseAsString();
```
rushtong: These two methods throw an IOException. If we hit those, we really should log them as errors. I think we could move these into the try/catch and deal with the different exceptions more explicitly. Lastly, if we did hit an error deserializing `search-terms.json`, we should log that as well since it is a different category of error.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/TranslateResource.java
```diff
@@ -1,56 +1,80 @@
...
-                    build();
+    public Response translateParagraph(final String jsonString) throws Exception {
+        final Map<String, String> body = mapper.readValue(jsonString, new TypeReference<>() {});
+        String paragraph = body.get("paragraph");
+
+        if (StringUtils.isBlank(paragraph)) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("Paragraph is required").build();
         }
+
+        return buildResponse(PARAGRAPH, paragraph);
```
rushtong: If you look at the other patterns in this class, we wrap the return in a try so we can control the exact response and log errors appropriately. Something like this would be more informative to both the client and devs reading logs

```
    try {
      return buildResponse(PARAGRAPH, paragraph);
    } catch (Exception e) {
      log.error("Error while translating paragraph: ", e);
      return Response.
          status(Response.Status.INTERNAL_SERVER_ERROR).
          entity("Error while translating paragraph: " + e.getMessage()).
          build();
    }
```

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -34,26 +43,41 @@ public class TextTranslationServiceImplTest extends AbstractTest {
     @Mock
     private AutocompleteService autocompleteService;
 
-    public TextTranslationServiceImplTest() {
+    @Mock
+    private GCSStore gcsStore;
+
+    @Mock
+    private StoreOntologyService storeOntologyService = getStorageServiceMock();
```
rushtong: This is unused.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -34,26 +43,41 @@ public class TextTranslationServiceImplTest extends AbstractTest {
...
+
+    public TextTranslationServiceImplTest() throws IOException {
```
rushtong: If you remove line 50, you can revert this line so it doesn't throw an exception

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -34,26 +43,41 @@ public class TextTranslationServiceImplTest extends AbstractTest {
...
+    }
+
+    private HttpResponse loadTermItemsResponse() throws IOException {
+      URL searchTerm = Resources.getResource("search-terms.json");
+      return getMockHttpResponse(Resources.toString(searchTerm, Charset.defaultCharset()));
+    }
+
+    private String getTitle(HashMap<String, Recommendation> translation, String urlKey) {
+      return translation.get(urlKey).title();
     }
```
rushtong: Minor nit, we should move this closer to the methods that use it, or to the end of the file where we often put private helper methods as opposed to having it float here in the middle of tests.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -89,6 +115,48 @@ public void testTranslateDataset() throws IOException {
         assertTrue(translation.contains("[GRU]"));
     }
 
+    @Test
+    public void testTranslateParagraphNormal() throws Exception {
+      // Given
```
rushtong, line 97: I like the given/when/then - reminds me of cucumber tests and some TDD ideas.

rushtong (changes requested):
See comments inline - this looks really good. Just some minor cleanup left.

src/main/java/org/broadinstitute/dsde/consent/ontology/model/TermItem.java
gabelmattos: I ran into the same issue. I decided to let it rest for now.

rushtong (approved):

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
```diff
@@ -1,17 +1,21 @@
...
 public interface TextTranslationService {
 
-    enum TranslateFor { DATASET, PURPOSE }
+    enum TranslateFor { DATASET, PURPOSE, PARAGRAPH }
 
     DataUseSummary translateDataUseSummary(String dataUseString);
 
     String translateDataset(String dataUse) throws IOException;
 
+    HashMap<String, Recommendation> translateParagraph(String paragraph) throws Exception;
```
pshapiro4broad: I think it's better to avoid implementation types like `HashMap` or `ArrayList` when possible, so I would use `Map` here.
```suggestion
    Map<String, Recommendation> translateParagraph(String paragraph) throws Exception;
```

Also, and I think I mentioned this before, I would avoid non-specific exception types like `Exception`.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
```diff
@@ -1,17 +1,21 @@
...
 import org.broadinstitute.dsde.consent.ontology.model.DataUseSummary;
+import org.broadinstitute.dsde.consent.ontology.model.Recommendation;
 
 import java.io.IOException;
+import java.util.HashMap;
 
 public interface TextTranslationService {
 
-    enum TranslateFor { DATASET, PURPOSE }
+    enum TranslateFor { DATASET, PURPOSE, PARAGRAPH }
```
pshapiro4broad, line 13: It seems odd to define an enum in an interface that doesn't use the enum. Is this missing a method that uses this type? Or maybe this type shouldn't be in this interface?

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,70 @@ public String translateDataset(String dataUseString) throws IOException {
...
+
+        for (String keyword : keywords) {
+          final boolean foundMatch = searchForKeyword(keyword, paragraph);
+
+          if (foundMatch) {
+            Recommendation recommendation = new Recommendation(
+                title,
+                category
+            );
+            if (!recommendations.containsKey(url)) {
```
pshapiro4broad: Since you only allow a single match for a `url`, you can `break` if `foundMatch` is `true` and avoid checking the rest of the keywords.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,70 @@ public String translateDataset(String dataUseString) throws IOException {
...
+              recommendations.put(url, recommendation);
+            }
```
pshapiro4broad: Since `if (!containsKey) { put(...); }` is a common form there's a java method for it, `putIfAbsent`. You can write this code as
```suggestion
            recommendations.putIfAbsent(url, new Recommendation(title, category));
```

If you want to avoid making a new object in the case where the map already contains a key, you can use `computeIfAbsent`
```suggestion
            recommendations.computeIfAbsent(url, key -> new Recommendation(title, category));
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,70 @@ public String translateDataset(String dataUseString) throws IOException {
...
+      try {
+        HttpResponse response = gcsStore.getStorageDocument("ontology/search-terms.json");
+        String terms = response.parseAsString();
+        return new Gson().fromJson(
+            terms,
+            new TypeToken<List<TermItem>>() {
+            }.getType()
+        );
+      } catch (Exception e) {
+        log.error("Error processing search terms from GCS: " + e.getMessage());
```
pshapiro4broad: Logging without an exception instead of throwing causes the exception's stack crawl to be lost. In this case I would rethrow using the caught exception, since you're going to throw anyway in the caller. That way the caller doesn't need to check for `isEmpty()` anymore.
```suggestion
        throw new Exception("Error processing search terms from GCS: ", e);
```

If you really don't want to rethrow, I would at least log the exception itself so the stack crawl for the error shows up in the logs.
```suggestion
        log.error("Error processing search terms from GCS: ", e);
```

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
```diff
@@ -196,4 +243,23 @@ public void testInvalidDate() throws IOException {
...
     }
 
+  /** Helper Methods **/
+
+  private void initializeTerm() throws Exception {
+      TermResource mockTerm = new TermResource();
+      mockTerm.setId("term id");
+      mockTerm.setLabel("term label");
+      mockTerm.setDefinition("term definition");
+      when(autocompleteService.lookupById(any())).thenReturn(Collections.singletonList(mockTerm));
```
pshapiro4broad: All the Collections singleton forms now (in JDK 17) have forms defined directly on the types, which I think are more convenient. For example, `List.of()` can be used without any arguments for an empty list, or with any number of arguments for a constant list. In this case:
```suggestion
      when(autocompleteService.lookupById(any())).thenReturn(List.of(mockTerm));
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
rushtong: Thank you, yes, will do some refactoring here.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
rushtong: We're using the enum in the resource class that does use this service. The `PARAGRAPH` case is a bit of an outlier, but it really is a kind of translation in the same vein as dataset and purpose translations.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: I like option 2, thanks!

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java
rushtong: We've been using List.of() for some time ... missed this one!

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
pshapiro4broad: It's used in the service implementation class but not in the service interface class, so it could be moved to the implementation class without changing the interface. Or maybe an impl method should be moved up to the interface, such as `translate(DataUse dataUse, TranslateFor translateFor)`? Then it would be used in both.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,67 @@ public String translateDataset(String dataUseString) throws IOException {
         return translate(dataUse, TranslateFor.DATASET);
     }
 
+    @Override
+    public Map<String, Recommendation> translateParagraph(String paragraph) throws IOException {
+        return paragraph(paragraph);
+    }
```
connorlbark: just curious from a design perspective: why separate `translateParagraph` from its implementation at `paragraph`?

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,67 @@ public String translateDataset(String dataUseString) throws IOException {
...
+
+    private List<TermItem> loadTermsFromGoogleStorage() {
+      try {
+        HttpResponse response = gcsStore.getStorageDocument("ontology/search-terms.json");
+        String terms = response.parseAsString();
+        return new Gson().fromJson(
+            terms,
+            new TypeToken<List<TermItem>>() {
+            }.getType()
```
connorlbark, line 150: would this be constant? creating a new object here seems odd to me.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
```diff
@@ -89,13 +98,67 @@ public String translateDataset(String dataUseString) throws IOException {
...
+            break;
+          }
+        }
+      }
+
+      return recommendations;
+    }
+
+    private static boolean searchForKeyword(final String keyword, final String targetText) {
+      return StringUtils.containsIgnoreCase(targetText, keyword);
```
connorlbark, line 140: does this mean that if the paragraph contained, for example, the word `collection`, it would be tagged as containing `col`? Should this use regex to ensure that the keyword is surrounded by whitespace?

connorlbark (commented):
couple comments ! looks very very solid

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationService.java
rushtong: > Or maybe an impl method should be moved up to the interface, such as `translate(DataUse dataUse, TranslateFor translateFor)`? Then it would be used in both.

I like that ... keeps things consistent between the interface and the implementation.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: This is Gabe's original implementation, so I can't speak for him. I don't think it adds any clarity aside from pulling out a larger block of code. 6 of one, half-dozen of the other.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: @connorlbark - Yes,`keyword`: `col1` with `targetText`: `collection` would return true here.
@gabedemattos - I defer to you for the logical implementation. Is the intention here to find entire `keyword` word (with boundaries) within the target text or is it OK to to look for containing strings?

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
rushtong: `new TypeToken<List<TermItem>>() {}.getType()` is the type-specific way of getting a typed list class instead of just a generic list class. 

Here's another way to think of what this code is doing but doesn't work:
```
new Gson().fromJson(terms, List<TermItem>.class)
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
gabelmattos: I'm not fond of going with REGEX for this. In a future implementation we will switch to ElasticSearch, this is just a prototypical solution.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java
gabelmattos: Great point, we merged a segregated file into this one and I didn't realize it was the case. 

Thanks.

quazi-h (approved):
This new endpoint and logic to recommended Data Use Ontology terms from the provided paragraph looks good to me. I did have one question - I was expecting this to be a GET endpoint since that's what specified in the ticket and we aren't really writing any data but is POST being used here because it's a function that is doing something (translating)?

PR #742, opened 2023-01-06
[DUOS-2154][risk=no] Custom Error Resource

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2154

See also: https://github.com/DataBiosphere/consent/pull/1872

### New 404 message
Screen Shot 2023-01-06 at 2 55 49 PM

### Old 404 message
Screen Shot 2023-01-06 at 2 55 31 PM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
@@ -1,0 +1,25 @@
+package org.broadinstitute.dsde.consent.ontology.resources;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import org.broadinstitute.dsde.consent.ontology.model.Error;
+import org.eclipse.jetty.server.Request;
+
+@Path("error")
+public class ErrorResource {
+
+  @GET
+  @Path("404")
+  @Produces("application/json")
+  public Response notFound(@Context HttpServletRequest request) {
+    String originalUri = ((Request) request).getOriginalURI();
+    String msg = String.format("Unable to find requested path: '%s'", originalUri);
+    Error error = new Error(msg, 404);
+    return Response.status(error.getCode()).entity(error).build();
+  }
+
+}
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.consent.ontology.resources;
 
+import java.net.URLDecoder;
+import java.nio.charset.Charset;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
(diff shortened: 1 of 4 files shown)


quazi-h (approved):
Changes look good to me, the error message is in line with what consent is returning.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
```diff
@@ -0,0 +1,25 @@
...
+import org.eclipse.jetty.server.Request;
+
+@Path("error")
+public class ErrorResource {
+
+  @GET
+  @Path("404")
+  @Produces("application/json")
+  public Response notFound(@Context HttpServletRequest request) {
+    String originalUri = ((Request) request).getOriginalURI();
```
connorlbark, line 21: I believe you can avoid this request cast by doing `@Context Request request` in the parameters. Some googling found you could also do

```
@Context
private final @Context Request request;
 ```

As a field. I don't know if this function will expand in size, but I would imagine if it did they would all need the request object, so that could be a way to expose it to all at once.

File: src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+    String msg = String.format("Unable to find requested path: '%s'", originalUri);
```
connorlbark: There must be something weird happening with this unicode decode: 
image

connorlbark (commented):
Looking great, couple things to cleanup then 👍

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
rushtong: When I tried that approach, the app choked on trying to cast the incoming `HttpServletRequest` as a `Request`. I'll have a second look and see if there's a cleaner way to avoid the cast.

rushtong: We are limited in what [we can pull in via `@Context`](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/jaxrs-resources.html#d0e2871):

> 3.6. Use of @Context
>
> Previous sections have introduced the use of [@Context](https://jakartaee.github.io/rest/apidocs/2.1.6/javax/ws/rs/core/Context.html). Chapter "Context" in the JAX-RS specification presents all the standard JAX-RS Java types that may be used with [@Context](https://jakartaee.github.io/rest/apidocs/2.1.6/javax/ws/rs/core/Context.html).
> 
> When deploying a JAX-RS application using servlet then [ServletConfig](http://docs.oracle.com/javaee/5/api/javax/servlet/ServletConfig.html), [ServletContext](http://docs.oracle.com/javaee/5/api/javax/servlet/ServletContext.html), [HttpServletRequest](http://docs.oracle.com/javaee/5/api/javax/servlet/http/HttpServletRequest.html) and [HttpServletResponse](http://docs.oracle.com/javaee/5/api/javax/servlet/http/HttpServletResponse.html) are available using [@Context](https://jakartaee.github.io/rest/apidocs/2.1.6/javax/ws/rs/core/Context.html).

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
rushtong: That's url encoding happening inside `getOriginalURI` ... gonna see if I can make that prettier.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/ErrorResource.java
connorlbark: Very interesting. Ok, I am okay with the cast. I saw examples of using `@Context` to pull Request... but perhaps that is in other versions.

connorlbark (approved):
👍 looks good

PR #778, opened 2023-02-21
[DUOS-2350]Data-use-v3

kaitlynoflaherty:
## Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2350

Creating a new data use schema for algorithm v3

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
@@ -20,13 +20,19 @@ import org.parboiled.common.FileUtils;
 public class DataUseResource implements OntologyLogger {
 
     @GET
-    @Path("/data-use-v3")
+    @Path("/data-use")
     @Produces(MediaType.APPLICATION_JSON)
     public Response getSchema() {
+        String content = FileUtils.readAllTextFromResource("data-use.json");
+        return Response.ok().entity(content).type(MediaType.APPLICATION_JSON).build();
+    }
+    @GET
+    @Path("/data-use-v3")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getSchemaV3() {
         String content = FileUtils.readAllTextFromResource("data-use-v3.json");
         return Response.ok().entity(content).type(MediaType.APPLICATION_JSON).build();
     }
-
     /**
      * This endpoint will take any valid form of json string that models DataUse and generate a Consent
      * UseRestriction using current business rules.
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
@@ -20,10 +20,10 @@ import org.parboiled.common.FileUtils;
 public class DataUseResource implements OntologyLogger {
 
     @GET
-    @Path("/data-use")
+    @Path("/data-use-v3")
     @Produces(MediaType.APPLICATION_JSON)
     public Response getSchema() {
-        String content = FileUtils.readAllTextFromResource("data-use.json");
+        String content = FileUtils.readAllTextFromResource("data-use-v3.json");
         return Response.ok().entity(content).type(MediaType.APPLICATION_JSON).build();
     }
 
--- a/src/main/resources/data-use-v3.json
+++ b/src/main/resources/data-use-v3.json
@@ -37,6 +37,9 @@
     "commercialUse": {
       "type": "boolean"
     },
+    "NPU": {
+      "type": "boolean"
+    },
     "other": {
       "type": "string"
     },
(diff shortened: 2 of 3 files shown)


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
```diff
@@ -20,10 +20,10 @@
...
 
     @GET
-    @Path("/data-use")
+    @Path("/data-use-v3")
     @Produces(MediaType.APPLICATION_JSON)
     public Response getSchema() {
-        String content = FileUtils.readAllTextFromResource("data-use.json");
+        String content = FileUtils.readAllTextFromResource("data-use-v3.json");
         return Response.ok().entity(content).type(MediaType.APPLICATION_JSON).build();
     }
```
rushtong: We're going to want both endpoints here, so instead of changing the current one, just add a new one.

src/main/resources/data-use-v3.json
```diff
@@ -0,0 +1,83 @@
...
+  "title": "Data Use Schema V3",
+  "version": 2,
+  "type": "object",
+  "dependencies": {
+    "recontactingDataSubjects": {
+      "oneOf": [
+        {"required": ["recontactMay"]},
+        {"required": ["recontactMust"]}
+      ]
+    },
```
rushtong: Don't need this, AFAICT - it's not in https://github.com/DataBiosphere/consent/blob/develop/src/main/resources/dataset-registration-schema_v1.json

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/DataUseResource.java
```diff
@@ -26,7 +26,13 @@ public Response getSchema() {
...
-
+    @GET
+    @Path("/data-use-v3")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getSchemaV3() {
+        String content = FileUtils.readAllTextFromResource("data-use-v3.json");
+        return Response.ok().entity(content).type(MediaType.APPLICATION_JSON).build();
+    }
```
rushtong, line 35: Don't forget that we need an entry for this in api-docs.yaml

kaitlynoflaherty: Do I need to update everything with 'data-use' in the path to 'data-use-v3', or just add an additional path for ' /schemas/data-use-v3'?

rushtong: Just add a new path and leave the existing one in place

src/main/resources/data-use-v3.json
```diff
@@ -0,0 +1,66 @@
...
+    "diseaseRestrictions": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "uniqueItems": true
+      }
+    },
+    "hmbResearch": {
+      "type": "boolean"
+    },
```
rushtong, line 35: One thing we need to figure out here is how to represent the condition that at most one of these three are selected. Let's circle back with Jonathan on this question and review what we're doing on both the dataset and dar UI forms. This might be helpful: https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=oneof

rushtong: Sorry, that comment didn't capture the three: `generalUse`, `diseaseRestrictions`, or `hmbResearch`.

src/main/resources/data-use-v3.json
```diff
@@ -0,0 +1,66 @@
...
+    "populationOriginsAncestry": {
+      "type": "boolean"
+    },
+    "methodsResearch": {
+      "type": "boolean"
+    },
+    "commercialUse": {
+      "type": "boolean"
+    },
```
rushtong, line 44: I think I'd rather see an `NPU` case here. This has been a sticking point for a long time since we're not actually capturing non-profit use with this question. Alternatively, maybe we should capture `NPU` in addition to commercial. I can see these cases:
1. A dataset is available only for commercial use
2. A dataset is available only for non-profit use
3. A dataset is available for both non-profit and commercial use
Arguably, case 3 could be described with a `null` value in either case, but that is ambiguous.

src/main/resources/data-use-v3.json
```diff
@@ -0,0 +1,69 @@
...
+    "NPU": {
+      "type": "boolean"
+    },
```
rushtong: Super minor nit, we should probably case this similarly to the rest of the fields.

src/main/resources/data-use-v3.json
fboulnois: Do we need to encode the `oneOf` restriction before merging?

fboulnois (approved):
lgtm, one question:

rushtong (approved):
Thank you for the changes 👍🏽

PR #794, opened 2023-03-21
[DUOS-2069]Translate V3 Business Logic into Code

kaitlynoflaherty:
## Addresses 
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2069

----

- Verify all tests go green


src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,238 @@
...
+  private AutocompleteService autocompleteService;
+
+  @BeforeEach
+  public void setUp() {
+    openMocks(this);
+    try {
+      when(autocompleteService.lookupById(any())).thenReturn(Collections.emptyList());
+    } catch (Exception e) {
+      //
+    }
```
rushtong, line 39: nit: this is unnecessary, tests will pass without this condition.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,91 @@
...
+      purposeDiseaseIdMap = generatePurposeDiseaseIdMap(purpose.getDiseaseRestrictions());
+    } catch (Exception e) {
+      String purposeRestrictions = StringUtils.join(purpose.getDiseaseRestrictions(), ", ");
+      List<String> errors = Arrays.asList(e.getMessage(), "Error found in one of the purpose terms: " + purposeRestrictions);
+      return ImmutablePair.of(false, errors);
+    }
+
+    ImmutablePair<Boolean, List<String>> diseaseMatch = matchDiseases(purpose, dataset, purposeDiseaseIdMap);
+    final List<ImmutablePair<Boolean, List<String>>> matchReasons = new ArrayList<>();
+    matchReasons.add(matchDiseases(purpose, dataset, purposeDiseaseIdMap));
```
rushtong: Calling `matchDiseases` twice is a little inefficient - this should work ... 
```suggestion
    ImmutablePair<Boolean, List<String>> diseaseMatch = matchDiseases(purpose, dataset, purposeDiseaseIdMap);
    final List<ImmutablePair<Boolean, List<String>>> matchReasons = new ArrayList<>();
    matchReasons.add(diseaseMatch);
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,91 @@
...
+    matchReasons.add(matchHMB(purpose, dataset));
+    matchReasons.add(matchMDS(purpose, dataset, diseaseMatch.getLeft()));
+    matchReasons.add(matchPOA(purpose, dataset));
+    matchReasons.add(matchCommercial(purpose, dataset));
```
rushtong: organization nit: Something that would be helpful would be to try and align the order here with the doc so it's easier to follow along and ensure that we're covering all cases. Additionally, we should probably re-structure the doc so we're doing the checks in a more consistent order. I would suggest doing the primary cases first (gru, hmb, disease) then the secondary cases after that.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,230 @@
...
+   * Datasets:
+   *      Any dataset tagged with GRU
+   *      Any dataset tagged with HMB
+   *
+   * @param purpose The data use object representing the Research Purpose
+   * @param dataset The data use object representing the Dataset
+   */
+  static ImmutablePair<Boolean, List<String>> matchHMB(DataUseV3 purpose, DataUseV3 dataset) {
+    // short-circuit hmb if not set
+    if (purpose.getHmbResearch() == null && dataset.getHmbResearch() == null) {
```
rushtong: nit: Prefer `Objects.nonNull` or `Objects.isNull` over `== null`

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  @Test
```
rushtong, line 42: We need something to test abstentions. For example, this test generates a positive match, which is incorrect:
```
  @Test
  public void testAbstain() {
    DataUseV3 dataset = new DataUseBuilderV3().setGeneralUse(true).build();
    DataUseV3 purpose = new DataUseBuilderV3().setOther("Other").build();
    ImmutablePair<Boolean, List<String>> match = matchPurposeAndDataset(purpose, dataset);
    System.out.println(match);
  }
```
results in ...
```
(true,[])
```

In addition to this, I'm not sure our model will accommodate the abstain conditions we define in the algorithm doc. I think we could update the v3 schema for this. If we had a single field, `abstain` (or something like that), then when we're parsing the object that comes in from the resource class (not in this PR), then we could trigger that field if there were any unknown fields in the json.

rushtong (commented):
See comments inline. I think we need some more design thinking around the abstain concept.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
kaitlynoflaherty: I completely agree, thanks!

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
kaitlynoflaherty: I'll start working on it!

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,276 @@
...
+   * @param dataset The data use object representing the Dataset
+   * @param purposeDiseaseIdMap is a map of each purpose term id to a list of that term's parent term ids
+   */
+  static MatchResult matchDiseases(
+      DataUseV3 purpose, DataUseV3 dataset, Map<String, List<String>> purposeDiseaseIdMap) {
+    // short-circuit if no disease focused research
+    if (purpose.getDiseaseRestrictions().isEmpty() && dataset.getDiseaseRestrictions().isEmpty()) {
+      return MatchResult.from(MatchResultType.APPROVE, Collections.emptyList());
+    }
+    // short-circuit if dataset is GRU
```
rushtong: Minor suggestion for comments here and throughout. In general, we're short-circuiting logic checks in the cases where there are no logical conditions to check. In cases where there ARE logical conditions to check, we're no longer short-circuiting. Instead, we're following the logic as outlined in the logic table.
```suggestion
    // Approve all GRU conditions
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,104 @@
...
+    } catch (Exception e) {
+      String purposeRestrictions = StringUtils.join(purpose.getDiseaseRestrictions(), ", ");
+      List<String> errors = Arrays.asList(e.getMessage(), "Error found in one of the purpose terms: " + purposeRestrictions);
+      return MatchResult.from(MatchResultType.DENY, errors);
+    }
+
+    MatchResult diseaseMatch = matchDiseases(purpose, dataset, purposeDiseaseIdMap);
+    final List<MatchResult> matchReasons = new ArrayList<>();
+    matchReasons.add(diseaseMatch);
+    matchReasons.add(matchDiseases(purpose, dataset, purposeDiseaseIdMap));
```
rushtong: This adds the disease matches twice. Suggest that you remove line 52.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,104 @@
...
+    List<String> purposeTermIdList = autocompleteService.lookupById(purposeDiseaseId)
+        .stream()
+        .filter(Objects::nonNull)
+        .filter(t -> Objects.nonNull(t.getParents()) && !t.getParents().isEmpty())
+        .flatMap(t -> t.parents.stream())
+        .map(p -> p.id)
+        .collect(Collectors.toList());
+    purposeTermIdList.add(purposeDiseaseId);
+    return purposeTermIdList;
+  }
```
rushtong: These methods are duplicated here from `DataUseMatcher`. We should pull these out into a utility class so they can be re-used by both algorithms.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,276 @@
...
+      if (!match) {
+        failures.add(String.format(DS_F2,  entry.getKey()));
+      }
+    }
+
+    if (failures.isEmpty()) {
+      return MatchResult.from(MatchResultType.APPROVE, failures);
+    } else {
+      return MatchResult.from(MatchResultType.DENY, failures);
+    }
```
rushtong: Note the very similar return code here. To streamline this, I would rewrite it first like this ...
```
    return failures.isEmpty() ? 
      MatchResult.from(MatchResultType.APPROVE, failures) :
      MatchResult.from(MatchResultType.DENY, failures);
```

Then, looking at it a little more closely, the only thing that's difference is the result type so I might rewrite it again as 

```
    MatchResultType type = failures.isEmpty() ? MatchResultType.APPROVE : MatchResultType.DENY;
    return MatchResult.from(type, failures);
```

You could even take it one step further, but I think this makes the code a little less legible:

```
    return MatchResult.from(failures.isEmpty() ? MatchResultType.APPROVE : MatchResultType.DENY, failures);
```

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,286 @@
...
+    DataUseV3 purpose = new DataUseBuilderV3().setHmbResearch(true).build();
+    assertApprove(purpose, dataset);
+  }
+
+  @Test
+  public void testHMB_negative() {
+    DataUseV3 dataset = new DataUseBuilderV3().setHmbResearch(true).build();
+    DataUseV3 purpose = new DataUseBuilderV3().setGeneralUse(true).build();
+    assertDeny(purpose, dataset);
+  }
```
rushtong: There should be two more negative HMB purpose cases according to the algorithm doc:

> Deny if the dataset's Primary DUO terms are DS-, POA

I wrote this to convince myself that the logic was working as expected. I leave it to you to organize as you see fit.
```
  @Test
  public void testMatchHMB() {
    DataUseV3 purpose = new DataUseBuilderV3().setHmbResearch(true).build();
    DataUseV3 dataset = new DataUseBuilderV3().setHmbResearch(true).build();
    assertApprove(purpose, dataset);

    purpose = new DataUseBuilderV3().setHmbResearch(true).build();
    dataset = new DataUseBuilderV3().setGeneralUse(true).build();
    assertApprove(purpose, dataset);

    purpose = new DataUseBuilderV3().setHmbResearch(true).build();
    dataset = new DataUseBuilderV3().setDiseaseRestrictions(Collections.singletonList(cancerNode)).build();
    assertDeny(purpose, dataset);

    purpose = new DataUseBuilderV3().setHmbResearch(true).build();
    dataset = new DataUseBuilderV3().setPopulationOriginsAncestry(true).build();
    assertDeny(purpose, dataset);
  }
```

rushtong (commented):
Still reviewing, but I think there's enough here to work with in the meantime.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
kaitlynoflaherty: Thank you!

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,273 @@
...
+      return MatchResult.from(MatchResultType.DENY, Collections.singletonList(HMB_F2));
+    }
+  }
+
+  /**
+   * RP: Study population origins or ancestry
+   * Future use is limited to research involving a specific population [POA]
+   * Datasets:
+   *      Any dataset tagged with GRU
+   *      Any dataset tagged with POA
```
rushtong, line 133: There is a deny case on the doc: 
> Deny if the dataset's Primary DUO terms are DS-, HMB

And the tests work in that case, we should document that here.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,309 @@
...
+        .setMethodsResearch(true).build();
+    assertDeny(purpose, dataset);
+  }
+
+  @Test
+  public void testMDS_negative_case_2() {
+    DataUseV3 dataset = new DataUseBuilderV3().setHmbResearch(true).build();
+    DataUseV3 purpose = new DataUseBuilderV3().setMethodsResearch(true).build();
+    assertDeny(purpose, dataset);
+  }
```
rushtong: I'm confused on these negative cases. For a MDS DAR, the docs say:
> Approve if the dataset's Primary DUO terms are GRU, DS-, HMB, POA

Case 1 should be a MR purpose against a DS dataset and, if the doc is correct, should be approved.
Case 2 should also be true if the doc is correct.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,309 @@
...
+  @Test
+  public void testCommercial_positive_case_2() {
+    DataUseV3 dataset = new DataUseBuilderV3().setGeneralUse(true).build();
+    DataUseV3 purpose = new DataUseBuilderV3().setCommercialUse(true).build();
+    assertApprove(purpose, dataset);
+  }
+
+  @Test
+  public void testCommercial_positive_case_3() {
+    DataUseV3 dataset = new DataUseBuilderV3().setNonProfitUse(null).build();
```
rushtong: nit, I think everything is null by default so setting it to null is redundant

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,273 @@
...
+   *      Any dataset where POA = true
+   *      Any dataset where DS-X match
+   */
+  static MatchResult matchMDS(DataUseV3 purpose, DataUseV3 dataset, MatchResultType diseaseMatch) {
+    // short-circuit if no disease focused research
+    if (purpose.getDiseaseRestrictions().isEmpty() && dataset.getDiseaseRestrictions().isEmpty()) {
+      return MatchResult.from(MatchResultType.APPROVE, Collections.emptyList());
+    }
+
+    boolean purposeMDS = getNullableOrFalse(purpose.getMethodsResearch());
```
rushtong, line 189: I think there might be a logic error here. There are technically three cases, but I think null or false are the same as you've calculated here.

1. Purpose MDS is true - we need to check the cases as outlined in the doc.
2. Purpose MDS is false/null - we don't need to do any checks - we can short circuit this because it isn't relevant to any decision making.

rushtong (commented):
Looking better! I think we need to look at MDS again, not sure if we're capturing that case correctly.

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
kaitlynoflaherty: I made a mistake here, I accidentally used the POA cases for MDS. I'll correct that now!

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,72 @@
...
+    matchReasons.add(matchPOA(purpose, dataset));
+    matchReasons.add(matchMDS(purpose, dataset, diseaseMatch.getMatchResultType()));
+    matchReasons.add(matchCommercial(purpose, dataset));
+    matchReasons.add(abstainDecision(purpose, dataset, purposeDiseaseIdMap, diseaseMatch.getMatchResultType()));
+    final boolean allMatch = matchReasons.stream().
+        map(MatchResult::getMatchResultType).
+        allMatch(rt -> rt.equals(MatchResultType.APPROVE));
+    final boolean anyMatch = matchReasons.stream().
+        map(MatchResult::getMatchResultType).
+        anyMatch(rt -> rt.equals(MatchResultType.ABSTAIN));
```
rushtong: Naming nit: This is determining an `anyAbstain` case.

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java
```diff
@@ -0,0 +1,72 @@
...
+        collect(Collectors.toList());
+    // if all items match, decision is APPROVED
+    if (allMatch) {
+      return MatchResult.from(MatchResultType.APPROVE, reasons);
+    }
+    // if not, determine whether DENY or ABSTAIN
+    if (anyMatch){
+      return MatchResult.from(MatchResultType.ABSTAIN, reasons);
+    }
+    return MatchResult.from(MatchResultType.DENY, reasons);
```
rushtong: Minor nit. This ternary can probably be simplified:
```
    MatchResultType type = allMatch ? MatchResultType.APPROVE : 
      anyAbstain ? 
        MatchResultType.ABSTAIN :
        MatchResultType.DENY;
    return MatchResult.from(type, reasons);
```

rushtong: Thinking out loud here - this logic could also be in a method inside `MatchResultType`, i.e., `getMatchTypeFromTypes(List<MatchResultType> matchResultTypes)` ... or something similar to that in concept.

src/main/java/org/broadinstitute/dsde/consent/ontology/model/DataUseV3.java
```diff
@@ -218,63 +217,63 @@ public String toString() {
         sb.append(DataUseV3.class.getName()).append('@').append(Integer.toHexString(System.identityHashCode(this))).append('[');
         sb.append("generalUse");
         sb.append('=');
-        sb.append(((this.generalUse == null)?"<null>":this.generalUse));
+        sb.append((Objects.isNull(this.generalUse)?"<null>":this.generalUse));
```
rushtong, line 220: Just a generic note here, not a request for a change. This is better style from a dev perspective. This code was auto-generated, so if we ever re-generate it, these changes would be overwritten.

rushtong (approved):
This looks really good, thank you for the changes! Comments inline are optional - implement as you see fit 👍🏽

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV3.java
```diff
@@ -0,0 +1,307 @@
...
+
+public class DataUseMatchCasesV3 {
+  private static final String HMB_F1 = "The GRU Research Purpose does not match the HMB data use limitations.";
+  private static final String HMB_F2 = "The HMB Research Purpose does not match the Disease-Specific data use limitations.";
+  private static final String HMB_F3 = "The HMB Research Purpose does not match the POA data use limitations.";
+  private static final String DS_F2 = "The Disease-Specific: %s Research Purpose is not a valid subclass of the Disease-Specific data use limitations.";
+  private static final String MDS_F1 = "The Methods development Research Purpose does not match the Disease-Specific data use limitations.";
+  private static final String POA_F1 = "The Populations, Origins, Ancestry Research Purpose does not match the HMB or Disease-Specific data use limitation.";
+  private static final String NCU_F1 = "The Commercial Use Research Purpose does not match the No Commercial Use data use limitation.";
+  private static final String Abstain = "The Research Purpose does not result in DUOS Decision.";
```
fboulnois:
```suggestion
  private static final String ABSTAIN = "The Research Purpose does not result in DUOS Decision.";
```

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/MatchResultType.java
```diff
@@ -0,0 +1,23 @@
...
+  public static Boolean Deny(MatchResultType x) {
+    return x == DENY;
+  }
+
+  public static Boolean Abstain(MatchResultType x) { return x == ABSTAIN; }
+
+  @Override
+  public boolean test(MatchResultType matchResultType) {
+    return false;
+  }
```
fboulnois: This method looks a little strange. Perhaps a comment on why this is needed would be useful?

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
```diff
@@ -0,0 +1,276 @@
...
+    TermParent parent1 = new TermParent();
+    parent1.setId("http://purl.obolibrary.org/obo/DOID_3119");
+    TermParent parent2 = new TermParent();
+    parent2.setId("http://purl.obolibrary.org/obo/DOID_0050686");
+    TermParent parent3 = new TermParent();
+    parent3.setId("http://purl.obolibrary.org/obo/DOID_162");
+    TermParent parent4 = new TermParent();
+    parent4.setId("http://purl.obolibrary.org/obo/DOID_14566");
+    TermParent parent5 = new TermParent();
+    parent5.setId("http://purl.obolibrary.org/obo/DOID_4");
```
fboulnois, line 62: I wonder if we could make these constants at the top of the class to make it easier to follow in the tests.

fboulnois (commented):
A few suggestions:

src/main/java/org/broadinstitute/dsde/consent/ontology/datause/MatchResultType.java
kaitlynoflaherty: I was able to remove a few things, this was able to go! thanks for pointing it out!

src/test/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3Test.java
kaitlynoflaherty: I think moving them would make it a bit more complicated than necessary. I can see the repetition between the first two tests but those are also the only tests that use these constants.

fboulnois (approved):

PR #807, opened 2023-04-10
[DUOS-2413]Creating MatchV3 endpoint

kaitlynoflaherty:
## Addresses 
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2413

----

- Verify all tests go green


src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResourceV3.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.List;
+
+@Path("/match")
+@Consumes("application/json")
+@Produces("application/json")
+public class MatchResourceV3 {
```
rushtong: I don't think we need a new resource class for this - we can just add a new endpoint in the existing match resource.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -64,4 +78,42 @@ public Response matchDataUse(final DataUseMatchPair matchPair) {
...
+        List<String> failures = DataUseUtil.sanitize(matchResult.getMessage());
+        return Response
+            .ok()
+            .entity(ImmutableMap.of(
+                "result", match,
+                "matchPair", matchPair,
+                "failureReasons", failures))
+            .type(MediaType.APPLICATION_JSON)
+            .build();
+      }
```
rushtong: I think semgrep is complaining about using `matchPair` on line 101. Maybe if you had a string as input, sanitized that string, then deserialized it with Gson, it might resolve the complaint.

rushtong: I did a test and I was wrong, that doesn't solve the problem. I suggest adding `// nosemgrep` just before the return statement

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -11,21 +11,34 @@
...
 
     @Inject
     MatchResource(DataUseMatcher dataUseMatcher) {
         this.dataUseMatcher = dataUseMatcher;
     }
 
+    @Inject
+    MatchResource(DataUseMatcherV3 dataUseMatcherV3) {
+      this.dataUseMatcherV3 = dataUseMatcherV3;
+    }
```
rushtong: Can you merge `DataUseMatcherV3` into the other constructor? Having two constructors with injection will be problematic.

kaitlynoflaherty: Is it possible to remove inject completely?

rushtong: I think it can - have not tested in this module yet.

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -64,4 +73,43 @@ public Response matchDataUse(final DataUseMatchPair matchPair) {
...
+        MatchResult matchResult = dataUseMatcherV3.matchPurposeAndDatasetV3(purpose, dataset);
+        MatchResultType match = matchResult.getMatchResultType();
+        List<String> failures = matchResult.getMessage();
+        // nosemgrep
+        return Response
+            .ok()
+            .entity(ImmutableMap.of(
+                "result", match,
+                "matchPair", matchPair,
+                "failureReasons", failures))
```
connorlbark, line 104: nit: maybe this could be its own object? but it looks good now as is, too.

src/main/resources/assets/api-docs.yaml
```diff
@@ -161,6 +161,29 @@ paths:
...
+    post:
+      summary: Match DataUse Purpose and Consent V3
+      description: Determine if a Research Purpose and Consent logically match.
+      parameters:
+        - name: body
+          in: body
+          description: Research Purpose and Dataset DataUse pair.
+          required: true
+          schema:
+            $ref: '#/definitions/DataUseMatchPairV3'
```
connorlbark, line 174: I don't think this or the other ref exist in Swagger?

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResourceV3Test.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  public void testOK() {
+    initResource();
+    DataUseV3 purpose = new DataUseBuilderV3().setHmbResearch(true).build();
+    DataUseV3 dataset = new DataUseBuilderV3().setGeneralUse(true).build();
+    DataUseMatchPairV3 pair = new DataUseMatchPairV3(purpose, dataset);
+    Response response = resource.matchDataUseV3(pair);
+    assertEquals(200, response.getStatus());
```
connorlbark: might be worth checking that the right fields are being sent back - and that "approve" / "deny" / "abstain" are all rendered properly.

connorlbark (commented):
looking great! couple things then 👍

pom.xml
```diff
@@ -8,9 +8,9 @@
     <name>Consent Management: Ontology Services</name>
 
     <properties>
-        <dropwizard.version>2.1.6</dropwizard.version>
-        <gatling.highcharts.version>3.9.3</gatling.highcharts.version>
-        <gatling.plugin.version>4.3.1</gatling.plugin.version>
+        <dropwizard.version>2.1.5</dropwizard.version>
+        <gatling.highcharts.version>3.9.2</gatling.highcharts.version>
+        <gatling.plugin.version>4.3.0</gatling.plugin.version>
```
rushtong: Revert all changes in this file, your base branch needs to be updated to match dev

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -11,20 +10,30 @@
...
+
+    private DataUseMatcherV3 dataUseMatcherV3;
 
-    @Inject
     MatchResource(DataUseMatcher dataUseMatcher) {
         this.dataUseMatcher = dataUseMatcher;
     }
+    MatchResource(DataUseMatcherV3 dataUseMatcherV3) {
+      this.dataUseMatcherV3 = dataUseMatcherV3;
+    }
```
rushtong: We should have only one constructor:
```suggestion
    MatchResource(DataUseMatcher dataUseMatcher, DataUseMatcherV3 dataUseMatcherV3) {
      this.dataUseMatcher = dataUseMatcher;
      this.dataUseMatcherV3 = dataUseMatcherV3;
    }
```

src/main/resources/assets/api-docs.yaml
rushtong: It does not reference anything, so it is needed.

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResourceV3Test.java
rushtong: Agreed

src/test/java/org/broadinstitute/dsp/ontology/performance/scenarios/MatchingV3.java
```diff
@@ -0,0 +1,29 @@
...
+  private final String v3MatchPair = """
+      {
+        "purpose": {
+          "generalUse": true
+        },
+        "consent": {
+          "generalUse": true
+        }
+      }
+      """;
```
rushtong, line 22: This is fine as is. Another way to do this is to generate a v3 object and call `toString` on it.

rushtong (changes requested):
See comments inline, looking good but a few updates are necessary.

src/test/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResourceV3Test.java
rushtong: Actually, Im revisiting my assessment and saying that I disagree here. As a resource test, we are mocking the service level response, so we don't care what the mocked service level response is ... the resource method is effectively a pass-through, so testing that the mocked `APPROVED` is really `APPROVED` is only testing the mock and not the method under test.

connorlbark: I mean in how it renders to JSON - like the APPROVED is becoming "approved"

pom.xml
rushtong: This is still an issue, these changes will roll back updates from other PRs.

rushtong (commented):
Aside from the pom.xml (there should be no changes to that file), this looks good.

kaitlynoflaherty:
> Aside from the pom.xml (there should be no changes to that file), this looks good.

I just had to update jacoco to the newest version

src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java
```diff
@@ -64,4 +72,43 @@ public Response matchDataUse(final DataUseMatchPair matchPair) {
         }
     }
 
+  /**
+   * Most recent version of matching logic as implemented in FireCloud
```
rushtong: Nit, we don't need this comment, it was only true for V2

rushtong (approved):
One last comment inline, then merge as you see fit 👍🏽

fboulnois (approved):
This feature is in good shape.

PR #958, opened 2024-05-07
[DCJ-323][risk=no] Lower warning on invalid schemas

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-323

### Summary
Minor change to lower log level for invalid schemas. These errors do not need to go to Sentry

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/dsde/consent/ontology/util/JsonSchemaUtil.java
+++ b/src/main/java/org/broadinstitute/dsde/consent/ontology/util/JsonSchemaUtil.java
@@ -95,7 +95,7 @@ public class JsonSchemaUtil implements OntologyLogger {
       Set<ValidationMessage> messages = schema.validate(jsonSubject);
       return messages.stream().map(ValidationMessage::getMessage).toList();
     } catch (Exception e) {
-      logError("Unable to load the data use schema: " + e.getMessage());
+      logWarn("Unable to load the data use schema: " + e.getMessage());
       throw new BadRequestException("Invalid schema");
     }
   }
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
Why is it safe to discard these errors from sentry?

rushtong:
> Why is it safe to discard these errors from sentry?

Because we inform the client that what they provided as input is invalid. Sentry reporting is for actionable errors that we, as developers have to do something for. If a client sends us bad input, there's nothing we can do about that.

fboulnois:
> > Why is it safe to discard these errors from sentry?
> 
> Because we inform the client that what they provided as input is invalid. Sentry reporting is for actionable errors that we, as developers have to do something for. If a client sends us bad input, there's nothing we can do about that.

Are there any situations where it would be valuable to record bad input, e.g. for recording potentially malicious input?

rushtong:
> > > Why is it safe to discard these errors from sentry?
> > 
> > 
> > Because we inform the client that what they provided as input is invalid. Sentry reporting is for actionable errors that we, as developers have to do something for. If a client sends us bad input, there's nothing we can do about that.
> 
> Are there any situations where it would be valuable to record bad input, e.g. for recording potentially malicious input?

We still report the issue, but not in sentry, and instead in GCP logging, so we can still follow up if we want to. There are no data related consequences to malicious input. If there were a security issue related to JSON parsing, I think that would be a separate concern outside the scope of what these endpoints are trying to do.

aarohinadkarni (approved):
looks good 👍

PR #999, opened 2025-01-22
DT-921: Add Spotless to build process.

rushtong:
### Addresses
Addresses https://broadworkbench.atlassian.net/browse/DT-921 only for Consent-Ontology

### Summary
Adds spotless to the build process. PRs will potentially fail the [maven action](https://github.com/DataBiosphere/consent-ontology/actions/workflows/maven.yaml) if there are any formatting problems.

This PR is limited to these things:
1. Adding the spotless plugin
2. Update the checkout for specific actions so they fetch `origin/develop`. Necessary to apply the `ratchetFrom` configuration option for spotless.
3. Applying the results of `mvn spotless:apply`
4. Ignore `semgrep` warning from updating `StorageFactory`

There are some outstanding cleanup problems that we can address in a future PR/ticket, for example, there are some unused classes touched in this PR like `UseRestrictionKeys` and `ValidationResponse`. 

----

- Verify all tests go green


.github/workflows/build.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         persist-credentials: false
+        fetch-depth: 0
```
rjohanek, line 27: what does this do?

rjohanek (approved):
spotless changes look good, thanks!

.github/workflows/build.yaml
fboulnois: This downloads the current version of the repository instead of the entire repository history.

rushtong: This is what I described in step 2 - it fetches the full history of the repo so that the local checkout has access to the `origin/develop` branch. See https://github.com/actions/checkout for docs. Without this, spotless isn't able to apply the `rachetFrom` configuration.

fboulnois:
How does this interact with our checked-in codeStyles?

rushtong:
> How does this interact with our checked-in codeStyles?

It applies the `googleJavaFormat` style which is what we currently use.

rushtong:
Going to close this for the time being. We need to better clarify where this should sit in the local build/dev/deploy cycle and how to deal with potential conflicts with `.idea/codeStyles`

